PR #33, opened 2013-12-10
Fix email address not linked. Add unicode option.

JillElaine:
http://trac.symfony-project.org/ticket/10065
Fix email address with no leading space not linked.
Add option for unicode characters in email username


miniche:
Open 3 years ago. Still relevant or can we close it?

PR #52, opened 2014-03-27
Replace sfValidatorEmail with php builtin filter_var function

nederdirk:
As this fork of Symfony1 depends on PHP 5.3 or greater, it is a good idea to use the PHP builtin function. The original implementation is at least as good as the Symfony regular expression and improves over the PHP versions.


miniche:
Mmh @j0k3r or @GromNaN: What do you want to do with this PR?

GromNaN:
👎 For [this reason](#issuecomment-38837510).

PR #63, opened 2014-08-28
Fix error

dkorsak:
Cannot use object of type sfValidatorPass as array


jeromemacias:
Sorry, we cannot revert like this, this was a WANTED feature, not a bug.

PR #74, opened 2014-11-26
add more factory as service via the services.yml configuration

mpoiriert:
We have more services available by default that are configure via the service container so it make this more flexible and more standard.

Also I had a change the was suppose to be in pullrequest #60 but it seem's that it wasn't done properly (the pullrequest or the merge I don't know) and the commit got lost...

### DIFF
--- a/lib/config/config/services.yml
+++ b/lib/config/config/services.yml
@@ -6,14 +6,66 @@
 #
 all:
   services:
-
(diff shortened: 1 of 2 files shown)


miniche:
Still relevant as this PR has been opened 2 years ago and the origin repo does not exist anymore?

PR #79, opened 2015-03-24
Change order of $factories

whatyoubendoing:
- Allows users to use the service container across other factories, for
  example in the session storage factory

### DIFF
--- a/lib/config/sfFactoryConfigHandler.class.php
+++ b/lib/config/sfFactoryConfigHandler.class.php
@@ -41,7 +41,7 @@ class sfFactoryConfigHandler extends sfYamlConfigHandler
     $instances = array();
 
     // available list of factories
-    $factories = array('view_cache_manager', 'logger', 'i18n', 'controller', 'request', 'response', 'routing', 'storage', 'user', 'view_cache', 'mailer', 'service_container');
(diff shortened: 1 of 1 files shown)


GromNaN:
:+1: I'll test.

j0k3r:
@b3njamin could you rebase your branch against the master to avoid these extra commits? Thanks

b3njamin:
Sure - done :)

j0k3r:
@miniche could you test that?

lib/config/sfFactoryConfigHandler.class.php
```diff
@@ -41,7 +41,7 @@ public function execute($configFiles)
     $instances = array();
 
     // available list of factories
-    $factories = array('view_cache_manager', 'logger', 'i18n', 'controller', 'request', 'response', 'routing', 'storage', 'user', 'view_cache', 'mailer', 'service_container');
+    $factories = array('service_container', 'view_cache_manager', 'logger', 'i18n', 'controller', 'request', 'response', 'routing', 'storage', 'user', 'view_cache', 'mailer');
```
alquerci: The service container depends on `user` and `routing` (e.g; `sfContext::getServiceContainer()`).

For example try to get a `service` that depend on `user` inside the constructor of the `response` class.
You will caught that the `sfContext::getInstance()->getUser()` return `null`. 

---

Another topic but it would be good to provide at least `i18n` and `mailer` as service.

alquerci (commented):
The trade that we can do currently without a refactoring for factories it just to get the depending service outside of the constructor.

PR #86, opened 2015-05-06
Add Multiple file input

vchabot:
Hello,

I added a new input form, which is an extension from sfWidgetFormInputFile, to allow multiple option (as you can upload multiple files with a single input file in HTML5).

It comes with its own validator.

### DIFF
--- a/lib/widget/sfWidgetFormInputFileMulti.class.php
+++ b/lib/widget/sfWidgetFormInputFileMulti.class.php
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * sfWidgetFormInputFileMulti represents an upload HTML input tag with multiple option.
+ *
+ * @package    symfony
+ * @subpackage widget
+ * @author     Vincent Chabot <vchabot@groupe-exp.com>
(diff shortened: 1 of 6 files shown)


lib/widget/sfWidgetFormInputFileMulti.class.php
```diff
@@ -0,0 +1,45 @@
...
+   * @param  array  $errors      An array of errors for the field
+   *
+   * @return string An HTML tag string
+   *
+   * @see sfWidgetForm
+   */
+  public function render($name, $value = null, $attributes = array(), $errors = array())
+  {
+    $name .= $this->getOption('multiple') ? '[]' : '';
+    return $this->renderTag('input', array_merge(array('type' => $this->getOption('type'), 'name' => $name, 'value' => $value, 'multiple' => $this->getOption('multiple')), $attributes));
```
GromNaN:
```php
if ($this->getOption('multiple'))
{
  $name .= '[]';
  $attributes['multiple'] = $this->getOption('multiple');
}
return parent::render($name, $value, $attributes, $errors);
```

PR #93, opened 2015-06-15
Large XML file save optimization

filipc:
Extractor task was not efficient.
Attempting to extract large amount of strings was causing empty xml in effect, because of memory issue with method $dom->saveXML();

Currently save method is rewritten to more memory efficient way using XMLReader and XMLWriter.
Script is using less memory and there is no need to allocate huge amount of memory to generate XML file.


j0k3r:
This reprents too much (un-tested) change to built-in component.

And the way sfMessageSource* are built allow you to use a custom one.

Create `lib/i18n/sfMessageSource_XLIFF2.class.php` :

```php
class sfMessageSource_XLIFF2 extends sfMessageSource_XLIFF
{
    // ...
}
```


and then in your factories.yml:

```yaml
  i18n:
    class: sfI18N
    param:
      source: XLIFF2
```

filipc:
I changed my solution as suggested. 
New class is named sfMessageSource_XLIFF_Large. It extends sfMessageSource_XLIFF and overrides only save method.

To activate that I added this file to sfCoreAutoload because core libraries are not using autoloader mechanism.

If anyone want to enable this MessageSource should add to factories.yml source definion: XLIFF_Large

GromNaN:
I'd recommand that you keep this new class in your own project (or plugin).

PR #96, opened 2015-07-14
changed log level of sfException to CRITICAL (excluding 404 error)

issei-m:
As Symfony2, I want to log as CRITICAL instead of ERR for thrown exception (except 404 error).
What do you think?


GromNaN:
:+1: since this is consistent with [Symfony2+](https://github.com/symfony/symfony/blob/2aeaa22e3171b5e1f10f2cc2a7ff7f6608b65f9d/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php#L85)

PR #99, opened 2015-07-22
Revert "Makes sfAPCCache compatible with APCu"

GromNaN:
Reverts LExpress/symfony1#94

Ensure you get the apcu [version >= 4.0.7](https://pecl.php.net/package-changelog.php?package=APCu&release=4.0.7)


GromNaN:
APCu 4.0.7 is available on dotdeb since the last week only.
http://packages.dotdeb.org/pool/all/p/php5-apcu/

PR #113, opened 2016-01-15
Fix integers used in mt_rand()

ngse:
The first integer provided was larger than the second one which causes PHP to generate a warning. Have removed a 1 from the first integer to make it smaller than the second.

### DIFF
--- a/lib/task/generator/sfGenerateAppTask.class.php
+++ b/lib/task/generator/sfGenerateAppTask.class.php
@@ -118,7 +118,7 @@ EOF;
 
     if (true === $options['csrf-secret'])
     {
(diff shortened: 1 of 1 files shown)


j0k3r:
Wow, interesting ! Good catch. :+1:

j0k3r:
A 7 year old bug then :trollface: 
http://trac.symfony-project.org/changeset/17036/branches/1.3/lib/task/generator/sfGenerateAppTask.class.php

ngse:
Yeah I figured it'd probably been there... a while. First time I've encountered it with PHP Warnings enabled on console :dancers:

PR #115, opened 2016-01-29
Change parameter of sfAction::renderJson

zobzn:
Why sfAction::renderJson(array $data) and not sfAction::renderJson($data)?


GromNaN:
:+1:

PR #122, opened 2016-05-13
travis does not use --remote on git submodule update

zkiiito:
so, the tests are running with old versions of doctrine and swiftmailer:

```
Submodule path 'lib/plugins/sfDoctrinePlugin/lib/vendor/doctrine': checked out '3d95c7f506e8b29d8e1fd48573dd0d046b9e6ae6'
Submodule path 'lib/vendor/swiftmailer': checked out '0697e6aa65c83edf97bb0f23d8763f94e3f11421'
```
with this update:

```
Submodule path 'lib/plugins/sfDoctrinePlugin/lib/vendor/doctrine': checked out '7abe498f56d383fabcd92ccc87f5d1a72c47f7f9'
Submodule path 'lib/vendor/swiftmailer': checked out '1460e98b53d2f916f1b4123069c96c90844a1a4d'
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "lib/vendor/swiftmailer"]
 	path = lib/vendor/swiftmailer
 	url = https://github.com/swiftmailer/swiftmailer.git
+	branch = 5.x
 [submodule "lib/plugins/sfDoctrinePlugin/lib/vendor/doctrine"]
 	path = lib/plugins/sfDoctrinePlugin/lib/vendor/doctrine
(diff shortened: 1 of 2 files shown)


j0k3r:
I think it was mostly to avoid testing on updated submobules but rather always use the locked ones.

zkiiito:
yeah it could be, but now it does not test with the versions users pull from composer :(

j0k3r:
Ok it's fine for me.
Could you update the submodule too?

zkiiito:
it's done

j0k3r:
Thanks !

PR #124, opened 2016-05-13
sfMemcacheCache fix

### LINKED ISSUE #123: sfMemcacheCache fail
The tests on travis run without memcache enabled, so the sfMemcacheCacheTest just skips itself.
If I start the memcached server on travis & add the extension to php (5.x), they fail, [as you can see here](https://travis-ci.org/ingatlancom/symfony1/builds/130089150)

these 2 methods fail:
```
 /**
  * @see sfCache
  */
  public function get($key, $default = null)
  {
    $value = $this->memcache->get($this->getOption('prefix').$key);

    return false === $value ? $default : $value;
  }

  /**
   * @see sfCache
   */
  public function has($key)
  {
    return !(false === $this->memcache->get($this->getOption('prefix').$key));
  }
```
so, if we set('key', false), has('key') is also false, and get('key', 1) returns 1.
somehow it should save that a false key is stored.

zkiiito:
fixes #123


j0k3r:
Thanks for that :+1:

PR #126, opened 2016-05-26
Adds feature to declare optional services via @?service_name

fruit:
Allows to declare services as non-required dependency in `services.yml` file.

This feature is also implemented in [Symfony 2](http://symfony.com/doc/current/book/service_container.html#ignoring-missing-dependencies).

Example:

``` yaml
services:
  opt_args:
    class: ClassOptionalArguments
    arguments: ['@?bar']
```

The `@bar` service is missing in service container, but the usage of `opt_args` service will not throw the `InvalidArgumentException`.

``` php
$optArgsService = $container->getService('opt_args');
// exception is not thrown
```

### DIFF
--- a/lib/service/sfServiceContainerDumperPhp.class.php
+++ b/lib/service/sfServiceContainerDumperPhp.class.php
@@ -143,7 +143,22 @@ EOF;
     {
       if (is_object($callable[0]) && $callable[0] instanceof sfServiceReference)
       {
-        return sprintf("    %s->%s(\$instance);\n", $this->getServiceCall((string) $callable[0]), $callable[1]);
+        $refId = (string) $callable[0];
+
+        if (0 === strpos($refId, '?'))
+        {
+          return <<< BLOCK
+    \$configurator = {$this->getServiceCall($refId)};
+    if (null !== \$configurator)
+    {
+      \$configurator->{$callable[1]}(\$instance);
+    }
+
+BLOCK
+;
+        }
+
+        return sprintf("    %s->%s(\$instance);\n", $this->getServiceCall($refId), $callable[1]);
       }
       else
       {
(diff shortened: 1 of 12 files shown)


lib/service/sfServiceContainerDumperPhp.class.php
```diff
@@ -351,6 +368,10 @@ protected function getServiceCall($id)
       return '$this';
     }
 
+    if (0 === strpos($id, '?')) {
```
j0k3r: Could you move the `{` to the new line?

lib/service/sfServiceContainerDumperPhp.class.php
```diff
@@ -143,7 +143,24 @@ protected function addServiceConfigurator($id, $definition)
...
+      \$configurator->{$callable[1]}(\$instance);
+    }
+
+BLOCK
+;
+        }
+        else
+        {
+          return sprintf("    %s->%s(\$instance);\n", $this->getServiceCall($refId), $callable[1]);
+        }
```
j0k3r: This `else` isn't necessary, you can remove it and directly use the `return`.

fruit:
I've squeezed fixes into one single commit.

j0k3r:
Thanks! This looks very interesting.
Maybe @GromNaN can have a look too?

GromNaN:
I've never used such feature. 
The implementation looks correct.

fruit:
At the moment, the only known problem is that `sfServiceContainerDumperGraphviz` renders dependencies with a leading "?" as missing.

j0k3r:
@fruit do you plan to fix it then?

fruit:
@j0k3r I've fixed it, but did it without tests at the moment.

fruit:
All affected and new tests passes.

```bash
php test/unit/task/service/sfServicesGraphvizDumpTaskTest.php
find test/unit/service/ -name "*Test.php" -exec php {} \;
```

fruit:
Perhaps, it is worth adding a documentation about this feature and new task...

fruit:
@j0k3r @GromNaN any comments/questions/suggestions?

GromNaN:
[👍](https://github.com/LExpress/symfony1/pull/126#issuecomment-222483508)

PR #127, opened 2016-05-27
sfAutoload include fix for HHVM support

zkiiito:
fixes the 2 dubious tests, hhvm is not likely to fix this issue soon :(

### DIFF
--- a/lib/autoload/sfAutoload.class.php
+++ b/lib/autoload/sfAutoload.class.php
@@ -138,7 +138,15 @@ class sfAutoload
 
     $file = $configuration->getConfigCache()->checkConfig('config/autoload.yml');
 
-    $this->classes = include($file);
+    if (defined('HHVM_VERSION'))
+    {
+      // workaround for https://github.com/facebook/hhvm/issues/1447
+      $this->classes = eval(str_replace('<?php', '', file_get_contents($file)));
+    }
+    else
+    {
+      $this->classes = include $file;
+    }
 
     foreach ($this->overriden as $class => $path)
     {


j0k3r:
Linked to https://github.com/facebook/hhvm/issues/1447

j0k3r:
Now only APC is failing, that's great ! :+1:

lib/autoload/sfAutoload.class.php
```diff
@@ -138,7 +138,15 @@ public function reloadClasses($force = false)
 
     $file = $configuration->getConfigCache()->checkConfig('config/autoload.yml');
 
-    $this->classes = include($file);
+    if (defined('HHVM_VERSION'))
+    {
+      // workaround for https://github.com/facebook/hhvm/issues/1447
+      $this->classes = eval(str_replace('<?php', '', file_get_contents($file)));
```
j0k3r: Does `eval` is always enabled in HHVM ?

zkiiito: I do not know any other way to disable it than the [disable_functions](http://php.net/manual/en/ini.core.php#ini.disable-functions) ini directive.

j0k3r: And `disable_functions` doesn't seems to be part of HHVM for now: https://github.com/facebook/hhvm/issues/2745

zkiiito: as reading a bit further, eval is not a function, but a language construct, so it's not even possible to disable with disable_functions.

j0k3r:
Thanks, good job !

PR #128, opened 2016-05-31
sfDoctrineRecord::getDateTimeObject() returns current time when value is null

marekk:


lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecord.class.php
```diff
@@ -232,16 +237,20 @@ public function getDateTimeObject($dateFieldName)
    * @param DateTime $dateTimeObject    The DateTime instance to use to set the value
    * @return void
    */
-  public function setDateTimeObject($dateFieldName, DateTime $dateTimeObject)
+  public function setDateTimeObject($dateFieldName, DateTime $dateTimeObject = null)
   {
     $type = $this->getTable()->getTypeOf($dateFieldName);
     if ($type == 'date' || $type == 'timestamp' || $type == 'datetime')
     {
+      if (is_null($dateTimeObject))
```
jeromemacias: Please do not use `is_null()`, use `null === $...` instead.

lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecord.class.php
```diff
@@ -207,17 +207,22 @@ public function __call($method, $arguments)
   }
 
   /**
-   * Get the Doctrine date value as a PHP DateTime object
+   * Get the Doctrine date value as a PHP DateTime object, null if the value is not set
    *
    * @param string $dateFieldName   The field name to get the DateTime object for
-   * @return DateTime $dateTime     The instance of PHPs DateTime
+   * @return DateTime|null $dateTime     The instance of PHPs DateTime
```
miniche: Can you add an empty line before this ```@return``` tag? ([Like this](https://github.com/marekk/symfony1/blob/85b74b025b47af4af8bf7a304fdedb418016ebe1/lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecord.class.php#L135))

miniche:
Seems good for me. After the little CS fix, can you squash your commits in only one?

Ping @jeromemacias & @j0k3r for validation.

j0k3r:
:+1:

PR #131, opened 2016-06-30
[WIP] Replaces Symfony 1.5 DIC with Symfony 2/3 DIC

fruit:
_Admonition: It's not a pull request, the Work is In Progress. The reason of early demonstration is to get feedback, criticism and ideas._

It's an attempt, to use a Symfony's Component ["DependencyInjection"](http://symfony.com/doc/current/components/dependency_injection/index.html) in place of DIC implemented in Syfony 1.5.

Why I did it: 
- The old DIC does not allows to use [Tagged Services](https://symfony.com/doc/current/components/dependency_injection/tags.html)

Known disadvantages:
- not fully BC with old DIC 
 - The `constructor` keyword in old DIC is handled by `factory` keyword in new DIC
- at the moment minimal `PHP` version 5.5.9 (it's possible to use minimal PHP 5.3.3 by downgrading to [Symfony DIC: v.2.6.13](https://packagist.org/packages/symfony/dependency-injection#v2.6.13)
- The `sf_mailer` is declared almost with same configuration in two places (`services.yml` and `task_services.yml`)

Known advantages:
- we get all features implemented Symfony 2/3 DIC (e.g. auto-wiring, optional services, service properties, tagged services, compiler passes, parent dependency management and etc.)
- The Context's DIC separated from Task's DIC and can be handy enhanced just by added personal services into `services.yml` or `task_services.yml` files respectively.


composer.json
```diff
@@ -3,7 +3,11 @@
     "description": "Fork of symfony 1.4 with dic, form enhancements, latest swiftmailer and better performance",
     "license": "MIT",
     "require": {
-        "swiftmailer/swiftmailer": "~5.2"
+        "php": ">=5.5.9",
```
GromNaN: Upgrading the minimal version of PHP is not something that can be accepted for this repository, as it is focused on long term maintenance.

fruit:
It's clear. What if we create an evolving variant of Symfony 1.X.  Under a different git branch, or name it Symfony 1.6?

PR #132, opened 2016-07-15
Save i18n message.xml target attribute in CDATA section

ngscz:
If you want to save HTML tag in messages.xml <target> attribute. You need to add this code to <![CDATA[]]> 
With this update, <target> attribute is automatically created with <![CDATA[]]> section.
For example:
```
<trans-unit id="1">
   <source>My contacts</source>
   <target><![CDATA[]]></target>
</trans-unit>
```
If you want to add HTML tag to <target>, just need to add:
```
<trans-unit id="1">
   <source>My contacts</source>
   <target><![CDATA[My<br> Contacts]]></target>
</trans-unit>
```

### DIFF
--- a/lib/i18n/sfMessageSource_XLIFF.class.php
+++ b/lib/i18n/sfMessageSource_XLIFF.class.php
@@ -197,7 +197,7 @@ class sfMessageSource_XLIFF extends sfMessageSource_File
       $source = $dom->createElement('source');
       $source->appendChild($dom->createTextNode($message));
       $target = $dom->createElement('target');
-      $target->appendChild($dom->createTextNode(''));
+      $target->appendChild($dom->createCDATASection(''));
 
       $unit->appendChild($source);
       $unit->appendChild($target);
@@ -277,7 +277,8 @@ class sfMessageSource_XLIFF extends sfMessageSource_File
           // set the new translated string
           if ($node->nodeName == 'target')
           {
-            $node->nodeValue = $target;
+            $node->nodeValue = '';            
+            $node->appendChild($dom->createCDATASection($target));
             $targetted = true;
           }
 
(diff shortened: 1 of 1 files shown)


j0k3r:
Interesting.
Could you update `sfMessageSource_XLIFFTest` to test this new feature?

lib/i18n/extract/sfI18nApplicationExtract.class.php
```diff
@@ -45,6 +45,18 @@ public function extract()
...
+     $includeMorePaths = sfConfig::get('app_i18n-extract-additional_enabled'); 
+     if (isset($includeMorePaths) && ($includeMorePaths == true)) 
+     { 
+       $paths = sfConfig::get('app_i18n-extract-additional_paths'); 
+          
+       foreach ($paths as $path) 
+       { 
+         $this->extractFromPhpFiles($path); 
+       } 
+     }
```
alquerci: Out of scope modification but for sure it is an interesting feature.

lib/widget/sfWidgetFormInputHidden.class.php
```diff
@@ -32,5 +32,28 @@ protected function configure($options = array(), $attributes = array())
...
+    */
+  public function render($name, $value = null, $attributes = array(), $errors = array())
+  {
+     if ($this->getOption('multiple'))
+     {
+       $name .= '[]';
+       $attributes['multiple'] = $this->getOption('multiple');
+      }
+      return parent::render($name, $value, $attributes, $errors);
+  }
```
alquerci: Out of scope modification but for sure it is an interesting feature.

lib/i18n/sfMessageSource_XLIFF.class.php
```diff
@@ -197,7 +197,7 @@ public function save($catalogue = 'messages')
       $source = $dom->createElement('source');
```
alquerci: Why only put a CDATA for the target as for the source it can be good as well?

PR #142, opened 2016-12-04
Fixed yaml dumpArray() for PHP 7.1

martijnpieters:
The function `dumpArray()` returned warnings about non-numeric values in the inline function of the `array_reduce` call. This was fixed by adding an explicit check for a numeric value.

Run `php test/unit/yaml/sfYamlInlineTest.php` on PHP 7.1 to see the original problem and to verify my fix.

### DIFF
--- a/lib/yaml/sfYamlInline.class.php
+++ b/lib/yaml/sfYamlInline.class.php
@@ -134,7 +134,7 @@ class sfYamlInline
     if (
       (1 == count($keys) && '0' == $keys[0])
       ||
-      (count($keys) > 1 && array_reduce($keys, create_function('$v,$w', 'return (integer) $v + $w;'), 0) == count($keys) * (count($keys) - 1) / 2))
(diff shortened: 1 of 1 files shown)


j0k3r:
Looks like your change is breaking an other test: https://travis-ci.org/LExpress/symfony1/jobs/181123500
Could you please check what's could be wrong?

martijnpieters:
I don't think I've introduced that problem. The most recent test on PHP 7.0 on the master branch has te same issue with the IP validator: https://travis-ci.org/LExpress/symfony1/jobs/181278235

j0k3r:
Hum right, I'll check that and your PR.

martijnpieters:
Thanks for the quick response!

zobzn:
was:
array_reduce($keys, create_function('$v,$w', 'return (integer) $v + $w;'), 0)
maybe this is better in this case?
array_sum(array_map('intval', $keys))

martijnpieters:
That looks better indeed

j0k3r:
You can fire a PR for that :wink:

zobzn:
https://github.com/LExpress/symfony1/pull/146

PR #143, opened 2016-12-05
Fix some bugs with new PHP versions

j0k3r:
Looks like the validatorIp got some problem related to some PHP versions.
According to 3v4l.org behavior are quite different for reserved IPv4 224.0.0.0/4 https://3v4l.org/cAdWZ
Instead of keeping the failing test while PHP fix the problem, I've switched to 240.0.0.1 IP which is working.

I've also submitted a bug to PHP to see what could be wrong with 224.0.0.0 IPs.
https://bugs.php.net/bug.php?id=73653

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ php:
 matrix:
     fast_finish: true
     allow_failures:
-        - php: 7.1
         - php: hhvm
         - php: nightly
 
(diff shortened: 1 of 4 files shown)


j0k3r:
Also, looks like isn't not a bug.
224.0.0.0/4 isn't reserved anymore

PR #144, opened 2016-12-07
Add support of Throwables

akalongman:
Add support of Throwables, which introduced in PHP 7.
http://php.net/manual/en/class.throwable.php

Now `sfException` class works properly with any `Exception` and `Error` instances.
This PR does not have a backward compatibility breaking changes, so feel free to merge.

### DIFF
--- a/lib/exception/sfException.class.php
+++ b/lib/exception/sfException.class.php
@@ -10,9 +10,9 @@
  */
 
 /**
- * sfException is the base class for all symfony related exceptions and
+ * sfException is the base class for all symfony related throwables and
  * provides an additional method for printing up a detailed view of an
- * exception.
+ * throwable.
  *
  * @package    symfony
  * @subpackage exception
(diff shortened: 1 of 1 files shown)


akalongman:
Now `sfException` class perfectly works with errors like: Return value TypeError and etc.

lib/exception/sfException.class.php
```diff
@@ -244,6 +255,8 @@ static protected function outputStackTrace(Exception $exception)
 
       return;
     }
+    die('XXA');
```
j0k3r: Why?

lib/exception/sfException.class.php
```diff
@@ -101,20 +101,31 @@ public function printStackTrace()
...
+      try
+      {
+        $this->outputStackTrace($exception);
+      }
+      catch (Exception $e)
+      {
+      }
     }
 
+
```
j0k3r: This new line isn't required

lib/exception/sfException.class.php
```diff
@@ -101,20 +101,31 @@ public function printStackTrace()
...
       header('HTTP/1.0 500 Internal Server Error');
     }
 
-    try
-    {
-      $this->outputStackTrace($exception);
-    }
-    catch (Exception $e)
-    {
+    if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
```
j0k3r: Could you please fix the CS here?
The `{` should be on a new line.
Same for the `else`.

lib/exception/sfException.class.php
akalongman: Oh, sorry, my bad. I will remove. It was for debug purposes :)

lib/exception/sfException.class.php
akalongman: ok

lib/exception/sfException.class.php
akalongman: Sure

e1himself:
Ping @j0k3r 
Any reasons this is not merged?

j0k3r:
I missed it

e1himself:
Thanks :+1:

PR #149, opened 2017-01-27
sfDoctrinePlugin: Check if validator keys are not associative

nederdirk:
If the `sfValidatorDoctrineChoice` validator from sfDoctrinePlugin validates an associative array, the input is passed unqualified to `$query->andWhereIn()`, resulting in a PDO exception.

### DIFF
--- a/lib/plugins/sfDoctrinePlugin/lib/validator/sfValidatorDoctrineChoice.class.php
+++ b/lib/plugins/sfDoctrinePlugin/lib/validator/sfValidatorDoctrineChoice.class.php
@@ -46,6 +46,7 @@ class sfValidatorDoctrineChoice extends sfValidatorBase
 
     $this->addMessage('min', 'At least %min% values must be selected (%count% values selected).');
     $this->addMessage('max', 'At most %max% values must be selected (%count% values selected).');
+    $this->addMessage('invalid_keys', 'Array keys must be numeric and consecutive');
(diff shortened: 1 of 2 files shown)


lib/plugins/sfDoctrinePlugin/lib/validator/sfValidatorDoctrineChoice.class.php
```diff
@@ -76,6 +77,11 @@ protected function doClean($value)
 
       $count = count($value);
 
+      if (array_keys($value) !== range(0, $count-1))
```
alquerci: [CS] Arithmetic operators should be arounded by at least one space.

lib/plugins/sfDoctrinePlugin/lib/validator/sfValidatorDoctrineChoice.class.php
```diff
@@ -76,6 +77,11 @@ protected function doClean($value)
...
+      {
+        throw new sfValidatorError($this, 'invalid_keys', array('keys' => array_keys($value)));
```
alquerci: This is a technical bug and not an user one. Than I propose to just use `array_values()` to have a valid array for Doctrine.

Note: We need to be sure that the array is not empty because on this case all records will be fetched. Then is a performance issue there.

PR #153, opened 2017-03-03
Improve IDE support on symfony core code

e1himself:
- No code changes
- Absolutely BC
- Add return type hints comments
- Add class properties type hints comments
- Add method arguments type hints comments
- Fix inconsistency between doc comments and implementation (by fixing comments)


e1himself:
Anybody here? :)

j0k3r:
I'm here and afraid of checking all 107 changing files :fearful:

e1himself:
:) @j0k3r those are just comments. 
The changes are just sane descriptions of actual method behavior. 
For most of them you can tell they're good just by looking at the diff.

Here are two (pretty straighforward) examples:

lib/command/sfCommandApplication.class.php
```diff
@@ -18,19 +18,32 @@
...
+  /** @var sfTask */
+  protected $currentTask = null;
+  /** @var sfEventDispatcher */
+  protected $dispatcher = null;
+  /** @var array */
+  protected $options = array();
+  /** @var sfFormatter */
+  protected $formatter = null;
+  /** @var mixed */
+  protected $commandOptions;
```
GromNaN: Good catch. This attribute was not defined but dynamically assigned.

lib/controller/sfController.class.php
```diff
@@ -20,17 +20,20 @@
...
-    $renderMode        = sfView::RENDER_CLIENT,
-    $maxForwards       = 5;
-
+  /** @var sfContext */
+  protected $context = null;
+  /** @var sfEventDispatcher */
+  protected $dispatcher = null;
+  /** @var string[] */
+  protected $controllerClasses = array();
+  protected $renderMode = sfView::RENDER_CLIENT;
```
GromNaN: No PHP doc for these ?

lib/form/addon/sfFormObject.class.php
```diff
@@ -96,8 +96,8 @@ public function bindAndSave($taintedValues, $taintedFiles = null, $con = null)
 
     return false;
   }
-
-
+  
+
```
GromNaN: Useless whitespaces added here.

lib/generator/sfModelGeneratorConfiguration.class.php
```diff
@@ -464,6 +474,7 @@ public function getCredentials($action)
 
   public function getPager($model)
   {
+    // TODO: Probably `getPagerClass()` method should be abstract here. As well as `getPagerMaxPerPage`
```
GromNaN: Sure, the method is used but never required.

lib/controller/sfController.class.php
e1himself: Fixed.

lib/form/addon/sfFormObject.class.php
e1himself: Fixed with 99efd6e

e1himself:
@GromNaN 

> @return void is not necessary, because implicit. And it seems to be non-standard.

My arguments for leaving it in are:

1. It's more obvious to change `@return something` => `@return void` to emphasize that it's returning nothing.
2. There's ton of `@return void` notations throughout symfony code.
3. It *is* in standard. It's just optional. See https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html  
Also I think it will be even more commonly used after adopting PHP 7.1 with its `: void` return type hint.

----------------

Please check the changes and let me know if you think it's mergeable. Thanks :)

e1himself:
Ping @GromNaN

e1himself:
Ping @GromNaN @j0k3r

e1himself:
Thanks :smile:

PR #160, opened 2017-06-23
Fix sfEventLogger: add priority and event.name overriding

thePanz:


lib/log/sfEventLogger.class.php
```diff
@@ -22,13 +22,18 @@ public function initialize(sfEventDispatcher $dispatcher, $options = array())
     {
       $this->setLogLevel($this->options['level']);
     }
+
+    // Use the default "command.log" event if not overriden
+    if (!isset($this->options['event_name'])) {
+        $this->options['event_name'] = 'command.log';
```
j0k3r: 2 spaces instead of 4.

thePanz: Fixed, thanks for taking care of Symfony1 :)

PR #162, opened 2017-07-28
Don't try to autoload a trait if it's already been declared

kaime:
As of [version 1.5.7](https://github.com/LExpress/symfony1/blob/master/CHANGELOG.md#02022016-version-157), autoload of traits is supported. However, the `sfAutoload->loadClass()` method will fail if you're trying to autoload a trait that has been previously loaded. This won't happen with classes or interfaces, since that method will do nothing if the required class or interface is already registered (by calling `class_exists()` and `interface_exists()`).

This PR just calls `trait_exists()` too, so it won't fail when trying to load a trait that has been previously loaded.

### DIFF
--- a/lib/autoload/sfAutoload.class.php
+++ b/lib/autoload/sfAutoload.class.php
@@ -188,7 +188,7 @@ class sfAutoload
     $class = strtolower($class);
 
     // class already exists
-    if (class_exists($class, false) || interface_exists($class, false))
+    if (class_exists($class, false) || interface_exists($class, false) || trait_exists($class, false))
     {
       return true;
     }
--- a/lib/autoload/sfAutoload.class.php
+++ b/lib/autoload/sfAutoload.class.php
@@ -188,7 +188,7 @@ class sfAutoload
     $class = strtolower($class);
 
     // class already exists
-    if (class_exists($class, false) || interface_exists($class, false) || trait_exists($class, false))
+    if (class_exists($class, false) || interface_exists($class, false) || (function_exists('trait_exists') && trait_exists($class, false)))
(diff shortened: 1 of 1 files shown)


lib/autoload/sfAutoload.class.php
```diff
@@ -188,7 +188,7 @@ public function loadClass($class)
     $class = strtolower($class);
 
     // class already exists
-    if (class_exists($class, false) || interface_exists($class, false))
+    if (class_exists($class, false) || interface_exists($class, false) || trait_exists($class, false))
```
j0k3r: `trait_exists` does not exist in PHP 5.3

kaime: Oops.

We can check that with `function_exists()` or `PHP_VERSION_ID`, but that would mean the framework will behave differently on PHP 5.3 and 5.4+

Not checking trait has been already loaded while supporting autoloading them is inconsistent as well IMO, but I don't think there can be a workaround to make it work on PHP 5.3.

The only thing I can suggest is dropping support for PHP 5.3, which is 8 years old now and has been unsopported for over 3 years:

http://php.net/supported-versions.php

j0k3r: Since trait aren't supported in PHP 5.3 people trying to use trait on that version will get an error so I guess they won't use it.

It's ok for me if you use `function_exists`.

I won't drop PHP 5.3 mostly because symfony1 is already a legacy project that shouldn't be used for new project but only to maintain old project which needs some small improvment.

j0k3r:
Thanks!

PR #163, opened 2017-07-29
Fix/php 7.2 deprecation warnings

e1himself:
This pull replaces `create_function` usages to regular lambda functions (introduced in PHP 5.3) to improve PHP 7.2 compatibility. 

It will fix PHP nightly build compatibility at your Travis-CI output:


e1himself:
Hm... It means that part is not covered with tests. Alright. I'll fix that as well.

j0k3r:
Yeah, the test suite doesn't cover 100% of the code.
That's why green tests doesn't mean everything will be ok 😕

e1himself:
@j0k3r done!

PR #168, opened 2017-08-19
Fix a bug in the lime unit-testing lib

e1himself:
It's stupid noone did this before.

So I've noticed a bug in *lime* that hides a few bugs in the test suite.

### The problem

```php
$lime->is('Hello', true); // passes as green
$lime->is('Hello', 0); // passes as green
```

While it should fail. That produced a lot of test suite bugs (human mistakes and typos) that are hidden. 

**Example of a hidden bug**:

See [sfParameterHolderTest.php:38](https://github.com/LExpress/symfony1/blob/master/test/unit/util/sfParameterHolderTest.php#L38). Currently it passes. But it doesn't work in a way it's described in the test. Because its misusing a bitwise operation: ` &= `.

```php
$ref2 &= $ph->get('ref');  
// SAME AS: $ref2 = null & 'foobar'; 
// bitwise AND on a null and a string == 0

$ref2 &= 'barfoo'; 
// SAME AS: $ref2 = 0 & 'barfoo';
// bitwise AND on an int and a string == 0

$t->is($ref2 , $ref, '->get() returns a reference for the given key');
// ATTENTION:
// This is where a lime bug is hiding the test suite bug. 
// It compares 0 to 'foobar' and passes.

```

### The PR

- adds a unit test for *lime*  (green)
- adds a unit test for *lime* that reveals a buggy/unexpected behavior (red)
- fix the *lime* bug (fixes the test above, but reveals a bunch of other different test suite problems) (red)
- fix the revealed test suite bugs (green)

### DIFF
--- a/lib/vendor/lime/lime.php
+++ b/lib/vendor/lime/lime.php
@@ -200,17 +200,15 @@ class lime_test
   private function equals($exp1, $exp2)
   {
     if (is_object($exp1) || is_object($exp2)) {
-      $value = $exp1 === $exp2;
+      return $exp1 === $exp2;
     } else if (is_float($exp1) && is_float($exp2)) {
-      $value = abs($exp1 - $exp2) < self::EPSILON;
+      return abs($exp1 - $exp2) < self::EPSILON;
     } else if (is_string($exp1) && is_numeric($exp1) || is_string($exp2) && is_numeric($exp2)) {
       return $exp1 == $exp2;
     } else if (is_string($exp1) || is_string($exp2)) {
       return (string) $exp1 === (string) $exp2;
-    } else {
-      $value = $exp1 == $exp2;
     }
-    return $value;
+    return $exp1 == $exp2;
   }
 
   /**
--- a/lib/vendor/lime/lime.php
+++ b/lib/vendor/lime/lime.php
@@ -191,6 +191,29 @@ class lime_test
   }
 
   /**
+   * Compares two values and returns true if they are equal
+   *
+   * @param mixed  $exp1    left value
+   * @param mixed  $exp2    right value
+   * @return bool
+   */
+  private function equals($exp1, $exp2)
+  {
+    if (is_object($exp1) || is_object($exp2)) {
+      $value = $exp1 === $exp2;
+    } else if (is_float($exp1) && is_float($exp2)) {
+      $value = abs($exp1 - $exp2) < self::EPSILON;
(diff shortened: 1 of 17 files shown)


e1himself:
I don't undestand. The test is failing, but it seems it's using an outdated version of `lime.php`. :/

e1himself:
NVM. Travis was configured to cache lime.php. Fixed that with 6152eda

e1himself:
Ping @j0k3r

lib/vendor/lime/lime.php
```diff
@@ -191,6 +191,29 @@ public function ok($exp, $message = '')
...
+   * Compares two values and returns true if they are equal
+   *
+   * @param mixed  $exp1    left value
+   * @param mixed  $exp2    right value
+   * @return bool
+   */
+  private function equals($exp1, $exp2)
+  {
+    if (is_object($exp1) || is_object($exp2)) {
+      $value = $exp1 === $exp2;
```
j0k3r: Why not `return $exp1 === $exp2;` ? Same for others condition, you can directly return the value

e1himself: Because *lime* relies on `==` comparison. If we do `===` it will require to *fix* a lot more test cases.

j0k3r: Hum, you didn't get what I told.

Why not return early instead of using `$value`?

```php
if (is_object($exp1) || is_object($exp2)) {
  return $exp1 === $exp2;
} else if (is_float($exp1) && is_float($exp2)) {
  return abs($exp1 - $exp2) < self::EPSILON;
} else if (is_string($exp1) && is_numeric($exp1) || is_string($exp2) && is_numeric($exp2)) {
  return $exp1 == $exp2;
} else if (is_string($exp1) || is_string($exp2)) {
  return (string) $exp1 === (string) $exp2;
}

return $exp1 == $exp2;
```

e1himself: > a lot more test cases.

I'll check the impact and get back with numbers.

imiroslavov: j0k3r, done that way you can also get rid of the else's.

j0k3r: Yup

e1himself: Ah, I see. You're right. I think I just tried to do least possible changeset preserving the original `$value` approach. 

Let me change that.

e1himself: Done. Thanks for the heads up.

j0k3r:
Looking good!

e1himself:
Thanks! :+1:

PR #175, opened 2017-12-01
Do not activate the sfACPUCache if not supported

thePanz:
Deny to create an "inactive" sfAPCCache cache, when the apcu extension is not supported, instead
of creating the object that does not work as expected.

As for php7 the apcu extension requries the apcu-bc extension for backward compatibility reasons

### DIFF
--- a/lib/cache/sfAPCCache.class.php
+++ b/lib/cache/sfAPCCache.class.php
@@ -32,9 +32,15 @@ class sfAPCCache extends sfCache
    */
   public function initialize($options = array())
   {
-    parent::initialize($options);
+    // Check if the APC is enabled and the apc_store function is available
+    // Be aware that php7's apcu extension removes the  backward compatibility with the original
+    // apcu used in php 5.x. The "apcu-bc" extension to be enabled to provide backward compatibility.
+    if (!function_exists('apc_store') || !ini_get('apc.enabled'))
+    {
+      throw new sfConfigurationException('sfAPCCache class needs the "apc" extension and the "apc_store" function to be enabled.');
+    }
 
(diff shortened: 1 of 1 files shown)


lib/cache/sfAPCCache.class.php
```diff
@@ -32,9 +32,15 @@ class sfAPCCache extends sfCache
    */
   public function initialize($options = array())
   {
-    parent::initialize($options);
+    // Check if the APC is enabled and the apc_store function is available
+    // Be aware that php7's apcu extension removes the  backward compatibility with the original
+    // apcu used in php 5.x. The "apcu-bc" extension to be enabled to provide backward compatibility.
+    if (!function_exists('apc_store') || !ini_get('apc.enabled'))
+    {
+      throw new sfConfigurationException('sfAPCCache class needs the "apc" extension and the "apc_store" function to be enabled.');
```
alquerci: Throwing an exception here is a BC breaks. Then to provide a BC layer I propose to introduce an option for that.

lib/cache/sfAPCCache.class.php
```diff
@@ -32,9 +32,15 @@ class sfAPCCache extends sfCache
...
+    }
 
-    $this->enabled = function_exists('apc_store') && ini_get('apc.enabled');
```
alquerci: What do you think to take the opportunity to fix the detection whether it is enabled?

See https://github.com/symfony/symfony/pull/29041

thePanz:
Closing in favor of: https://github.com/FriendsOfSymfony1/CacheLegacyBridge

PR #178, opened 2017-12-12
Fix counting of non-countable var for PHP 7.2

vitaliytv:


lib/form/sfFormField.class.php
```diff
@@ -324,6 +324,6 @@ public function getError()
    */
   public function hasError()
   {
-    return null !== $this->error && count($this->error);
+    return null !== $this->error && (is_object($this->error) ? count($this->error) : false);
```
vendethiel: `null` isn't an object.

vitaliytv: see recent changes please

lib/form/sfFormField.class.php
```diff
@@ -324,6 +324,6 @@ public function getError()
    */
   public function hasError()
   {
-    return null !== $this->error && count($this->error);
+    return null !== $this->error && $this->error instanceof sfValidatorError;
```
GromNaN: An instance of a class cannot be null, so you can also remove `null !== $this->error`

e1himself: Hm.. I guess we might have a regression here.

Potentially `$this->error` can be an instance of `sfValidatorErrorSchema`, which is `Countable`.

So before `hasError()` did return *false* for *sfValidatorErrorSchema* with no nested errors.
After this change, it will return true for such cases. Bug?

e1himself: Yes, I think so. Please check: https://github.com/LExpress/symfony1/blob/master/lib/form/sfForm.class.php#L1184

This proofs an instance of `sfFormField` can be constructed with `sfValidatorErrorSchema`.

vendethiel: Yes it's missing the count now.
I think my comment was misinterpreted. I meant the same thing as @gromnan.

vitaliytv: return $this->error instanceof sfValidatorError || (($this->error instanceof sfValidatorErrorSchema) ? $this->error->count() : false); 
?

e1himself: @vitaliytv 

> ```php
> return $this->error instanceof sfValidatorError 
>     || (($this->error instanceof  sfValidatorErrorSchema) ? $this->error->count() : false);
> ```

The check will never execute the right part, as *sfValidatorErrorSchema* **is** an instance of *sfValidatorError* too :)

I suggest to be super explicit and write something like this:

```php
if ($this->error instanceof sfValidatorErrorSchema) {
    return $this->error->count() > 0;
}
return $this->error !== null;
```

vitaliytv:
test travis

vitaliytv:
Could you squash your commits? done

PR #179, opened 2018-01-17
Upgrade to swiftmail 6

hugochinchilla:


j0k3r:
I'm not sure what to do with that since breaking the compatibility will requiert to upgrade to a major version.
Do you really want a symfony1 2.0? :roll_eyes: 

If you really really need SwiftMailer 6 on your project, you should better look to move away from symfony1 first.

hugochinchilla:
You are right @j0k3r 

It was meant as an experiment, I was too quick making the PR, I didn't notice the fact that Swiftmailer requires php > 7.0 after seing the broken tests on travis.

PR #181, opened 2018-02-05
Fix error 'A non well formed numeric value' on sfValidatorFile

cjobeili:


cjobeili:
Yes, `upload_max_filesize` return 2M for example, so $max = 2M
And now we $max*1024 -> Non well formed numeric value.

cjobeili:
Yes no problem ! Thanks for this quick merge !

vendethiel:
That probably doesn't matter in practice, but if there's no suffix, it's gonna cut off the last digit.
i.e. if `upload_max_filesize` is `500`.

mentalstring:
As @vendethiel said, this effectively breaks the use of the regular syntax (vs [shorthand](http://php.net/manual/en/faq.using.php#faq.using.shorthandbytes)).

j0k3r:
@vendethiel @mentalstring thanks for watching!

@cjobeili could you update your fix so that it handle the regular syntax too?
Using 1024, it dumps 102 instead of 1024: https://3v4l.org/tSLBm

cjobeili:
#182 Here it is ! :)

PR #182, opened 2018-02-06
Fix usage of octet in recent patch

cjobeili:
Fix octet usage of #181 

Octet: https://3v4l.org/lHLUE
Short: https://3v4l.org/TbfAJ


j0k3r:
@vendethiel @mentalstring Are you ok with that extra fix?

vendethiel:
Seems good to me.

mentalstring:
The casting on the return is now unnecessary (ie: $value is casted twice to int), but this is nitpicking — it works fine like this.

j0k3r:
Thanks @cjobeili!

PR #184, opened 2018-02-27
fix for i18n bug on php7.1 returns false on isset evaluation...

jomiojeda:
…making translation return always default culture translation. It should be true.

### DIFF
--- a/lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecordI18nFilter.class.php
+++ b/lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecordI18nFilter.class.php
@@ -51,7 +51,7 @@ class sfDoctrineRecordI18nFilter extends Doctrine_Record_Filter
   public function filterGet(Doctrine_Record $record, $name)
   {
     $culture = sfDoctrineRecord::getDefaultCulture();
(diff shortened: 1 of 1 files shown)


jomiojeda:
I will try to when I have some time. For now the fix solved the problem for our client. Thanks.

alquerci:
It is not a PHP-7 bug but just a modification of the deep isset behaviour.

Deep isset() on class that implement ArrayAccess
----------------------------

```php
<?php

isset($arrayAccess['foo']['bar']);
```

### Before PHP-7.0

```php
<?php

$arrayAccess->offsetGet('foo')->offsetExists('bar');
```

### PHP-7.0 and after

```php
<?php

if ($arrayAccess->offsetExists('foo')) {
    $arrayAccess->offsetGet('foo')->offsetExists('bar');
}
```

Solution
----------

Just need to do the following modification.

```php
<?php

// The relationship must be fetched in order to check the culture existance.
// Related to PHP-7.0 compatibility so an explicit call to method get is required.
$translation = $record['Translation'];

// process the translation.
```

dr0bz (approved):
The fix works! 

What is missing now to get into master?

P.S.: here some more info on the isset & ArrayAccess php bug: https://bugs.php.net/bug.php?id=69659

alquerci (changes requested):

PR #185, opened 2018-02-28
Allow PATCH as an option in functional tests

mkopinsky:
We have some (Vue.js) app code where we use PATCH as a request method, and we needed to make this tweak to sfTestFunctionalBase to be able to test the routes accepting PATCH.


j0k3r:
Thanks for explaining 👍

Issue #186, opened 2018-03-01
PHP Fatal error:  Uncaught Error: Call to undefined function echoln()


antony-develop:
I'm trying to update symfony 1 on my project.
And i'm getting this error
`PHP Fatal error:  Uncaught Error: Call to undefined function echoln() in lib/services/MyClass.php`
There was the echoln() function in lib/yaml/sfYaml.php → lib/yaml/sfYaml.class.php file
It was removed in this commit
https://github.com/LExpress/symfony1/commit/10d535c80eec1bb22fd8b58ea2513f431c2e863b

Please put it back. I have a lot of places where this function is in use.

j0k3r:
Feel free to make a PR to re-add it

e1himself:
@antony-develop I think we should not bring back that function here.

1. It's not used inside this repository anymore.
2. If your project depends on it, I think it's better to just intoruduce it in your codebase.
3. The location feels so wrong: sfYaml class file is not a place for a global scope function.

cc @j0k3r

j0k3r:
Hell, I just merged the PR without checking back to the issue first.
And I agree with you @e1himself 😞

antony-develop:
@e1himself
Yes, I already introduce it in my code.
So I think you guys know better what to do with it.

e1himself:
@j0k3r should we have another PR to drop it again? :)

PR #187, opened 2018-03-02
add echoln function to lib/yaml/sfYaml.class.php

antony-develop:
As it was here before changes in this commit
https://github.com/LExpress/symfony1/commit/10d535c80eec1bb22fd8b58ea2513f431c2e863b


e1himself:
For the reference: #186

j0k3r:
Thanks @antony-develop!

PR #189, opened 2018-03-07
Fix case for isSubclassOf method

mkopinsky:
Similar to lexpress/doctrine1#47 (which addresses lexpress/doctrine1#40)


j0k3r:
Thanks @mkopinsky

PR #197, opened 2018-09-24
Changes for using memcached class (instad of old memcache)

JohannesTyra:
The old memache is not longer maintained.
In php 7.2 the package is not longer available.

Using the memcached class & updated the doctrine cache (memcache) also.

### DIFF
--- a/lib/cache/sfMemcacheCache.class.php
+++ b/lib/cache/sfMemcacheCache.class.php
@@ -19,14 +19,14 @@
 class sfMemcacheCache extends sfCache
 {
   /** @var Memcache */
-  protected $memcache = null;
+  protected $memcached = null;
 
   /**
    * Initializes this sfCache instance.
    *
    * Available options:
    *
-   * * memcache: A memcache object (optional)
+   * * memcache: A memcached object (optional)
    *
    * * host:       The default host (default to localhost)
    * * port:       The port for the default server (default to 11211)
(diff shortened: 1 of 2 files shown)


alquerci:
Hello @JohannesTyra ,

I suggest to create a dedicated `sfMemcachedCache` class in order to not introduce back compatibility breaks.

And I invite you to check [existing implementation on Symfony4][0]. The goal for sure is to not re-write the wheel.

Even more create a PSR-6 and PSR-16 adapter to `sfCache` will be more extendable (be careful with reserved characters for the cache key `{}()/\@:`).

[0]: https://github.com/symfony/symfony/blob/v4.1.4/src/Symfony/Component/Cache/Traits/MemcachedTrait.php

JohannesTyra:
I created a dedicated class / revert the changes in the old class.

e1himself:
@alquerci this is the time to close it, I guess.

alquerci:
@e1himself I think too. Moreover because `from braindata:master`.

cc @JohannesTyra

desyncr:
@alquerci @e1himself We have already a memcached implementation; back a couple of months ago I asked if it was OK to submit those changes; what's the current view on that?

alquerci:
> @alquerci @e1himself We have already a memcached implementation; back a couple of months ago I asked if it was OK to submit those changes; what's the current view on that?

@desyncr `Memcached` and `Memcache` are two distinct implementations of `memcached` client. As `Memcached` is like the next major version (with backward compatibility breaks). Both do not use the same PHP extension (if I am not wrong).

And as we focus on even more on fixes and help for Symfony2+ migration adding a new cache implementation is very low priority.

---

@thePanz As this PR took the `master` branch as remote. IMHO it is time to close it and open a new PR or at least an issue to keep track on this one.

desyncr:
@alquerci What I was trying to express was that at my company we have a `sfMemcachedCache` implementation (what is @JohannesTyra working on).

thePanz:
@desyncr do you have a pointer to that implementation?
@alquerci sorry, I sort of lost track of this mrs

thePanz:
@alquerci @desyncr : from my experience the sfApcCache is also not working anymore: the code from SF1 is using an old version of APC (not maintained). to function it is required to install the apcu-bc extension.
Worst thing: the sfApcCache class will silently fail to store/retrieve items if the PHP extensions, making hard to debug issues if the SfApcCache is used to store session data.

My proposal here: create a new repository that bridges the caches implemented in SF4 (as @alquerci suggested, see: https://github.com/symfony/symfony/blob/v4.1.4/src/Symfony/Component/Cache/Traits/MemcachedTrait.php)

thePanz:
@JohannesTyra what about providing the new cache layer in https://github.com/FriendsOfSymfony1/CacheLegacyBridge ?
IIRC SymfonyCache already supports the Memcached backend

thePanz:
@alquerci @JohannesTyra  @desyncr Is it OK for you to close this PR?

Issue #102, opened 2015-08-14
Friends of Symfony1 (FOS1) Organization?


thePanz:
From the idea on #100 .. what do  you think about merging our efforts to have a FriendsOfSymfony1 (FOS1) organization where let all the Symfony1 related works be unified?

Meaning also to have two different branches of Symfony 1.5.x (with BC-breaks) and Symfony 1.4.x (with fixes and backports only) I saw that other users proposed (and worked) on such 1.4.x updating without BC breaks here: https://github.com/punkave/symfony1 by @boutell

boutell:
I'm all for it as long as it's well maintained — there need to be enough
people who have commit privileges for the 1.4 branch that it doesn't go
stale, yet everyone doing so needs to firmly understand that there are
*zero* bc breaks there.

On Fri, Aug 14, 2015 at 5:48 AM, Ema Panz <notifications@github.com> wrote:

> From the idea on #100 <https://github.com/LExpress/symfony1/issues/100>
> .. what do you think about merging our efforts to have a FriendsOfSymfony1
> (FOS1) organization where let all the Symfony1 related works be unified?
>
> Meaning also to have two different branches of Symfony 1.5.x (with
> BC-breaks) and Symfony 1.4.x (with fixes and backports only) I saw that
> other users proposed (and worked) on such 1.4.x updating without BC breaks
> here: https://github.com/punkave/symfony1 by @boutell
> <https://github.com/boutell>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LExpress/symfony1/issues/102>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

thePanz:
Waiting for @j0k3r and @GromNaN feedback too.
Regarding the BC breaks: unit-tests (plus Travis integration?) could be used to ensure that the 1.4 branch do not get BC breaks. As far as I know the 1.5 branch is not fully successful with all the symfony1 tests.

thePanz:
[...]

PR #200, opened 2018-10-19
[WIP] Add ability to not only cache HTTP 200 and GET method

alquerci:
Add pages with response 200, 203, 300, 301, 302, 404, 410 to cache.
Add to cache response from HEAD method request.

TODO
----

- [ ] Add tests


alquerci:
@j0k3r @GromNaN Can you look for this patch?

I provided all required tests for it.

alquerci:
@GromNaN Thank you for your time and to take care of this.

alquerci:
Noted, then the goal will be to find a new repository based on this version and able to maintain using almost same quality as Symfony.

As I see on the discussion #102 there are many forks but we only need one.

The goal of this patch is to avoid adding another fork.

alquerci:
@thePanz @mkopinsky Can you take a look at this patch?

e1himself:
My point of view on pushing this repo further is to remove code instead of writing more. View caching, for example, should be completely unloaded from the framework to a specialized tool like Varnish.

alquerci:
Hello @e1himself 

Indeed, but I propose this patch because it fix a business issue.

alquerci:
@e1himself What do you think if we add a configuration parameter to activate this patch (to minimize any BC breaks)?

Sometimes a bug can be a feature.

e1himself:
> What do you think if we add a configuration parameter to activate this patch (to minimize any BC breaks)?

If you ask me, I'd say this fix should live in your codebase as a set of stock framework classes overrides. You can do that with `factories.yml` (even though it'll require quite lots of overrides).

I believe we should not bring new features into this repo.

alquerci:
@e1himself Thank you for your feedback.

This PR exists because there is no way to provide this patch as a set of stock framework classes overrides.

This PR provide a bug fix for [HTTP spec](https://tools.ietf.org/html/rfc7231#section-6.1).

Your message is a bit confusing as you start to discuss about a fix, but you finish your message as a feature.

lib/filter/sfExecutionFilter.class.php
```diff
@@ -90,7 +90,29 @@ protected function executeAction($actionInstance)
   {
     // execute the action
     $actionInstance->preExecute();
-    $viewName = $actionInstance->execute($this->context->getRequest());
+
+    try {
+      $viewName = $actionInstance->execute($this->context->getRequest());
```
alquerci: Why to not wrap pre and post execute methods?

e1himself:
@alquerci 

Hey Alexandre,

> This PR exists because there is no way to provide this patch as a set of stock framework classes overrides.

I might be wrong, but I see nothing preventing this.

> Your message is a bit confusing as you start to discuss about a fix, but you finish your message as a feature.

Yes, you're right. I've used inaccurate wording :)
Technically this is a bug fix. But I wanted to say that this PR is bringing something new. Even if it's a missing aspect of an existing feature.

Again, it's not me who should decide on whether this should be merged or not :)

alquerci:
> Again, it's not me who should decide on whether this should be merged or not :)

@e1himself Indeed, but as open-source project the point of view of any code reviewer can be taken account.

As you shared your point of view it means that you are involved on the project. And just for that, thank you.

---

Aside you I am agreed with you about:
> View caching, for example, should be completely unloaded from the framework to a specialized tool like Varnish.

However for a project where there are more than 10 million of landing pages on cache. What is the better solution between :
  1. Migrate the fully cache system to Varnish or other system
  2. Contribute to the framework to provide the patch to share it to other users using the `view_cache`.
  3. Fork the framework and apply the patch, but closing door to open-source community (the company have not enough resources to maintain a custom version).

While solutions first and second are relatively reasonable the third is something to avoid if we want to have a useful @FriendsOfSymfony1 community (at least until the last sf1 project will migrate)

---

I think that such discussion is very constructive for the mindset of @FriendsOfSymfony1 community.

cc @thePanz

PR #202, opened 2018-12-06
Normalise min/max-options for sfValidatorDate

nederdirk:
The `min`/`max` options of `sfValidatorDate` were not normalised in the same way as the input, resulting in the following incorrect behaviour:
```php
$validator = new sfValidatorDateTime([
  'min' => '2018-12-06T09:00:00+00:00',
]);
$validator->clean('2018-12-06T09:59:59+01:00'); // this should throw, because 09:59+01:00 is one minute before 09:00+00:00, but it doesn't
```

This merge request adds tests for this situation and fixes this behaviour


lib/validator/sfValidatorDate.class.php
```diff
@@ -117,6 +117,7 @@ protected function doClean($value)
       else
       {
         $dateMax  = new DateTime($max);
+        $dateMax->setTimezone(new DateTimeZone(date_default_timezone_get()));
```
alquerci: Hello @nederdirk,

Here is not an issue with the time zone but the way where the date time are compared.

The proper way to do it is to compare timestamps.

As the timestamp does not depends on the timezone then all issue will be resolved there.

PR #205, opened 2019-02-09
Fix First translation access

pouletfou:
I ran into a problem where the first Translated field I tried to access would return in the DefaultCulture instead of the current culture because the $record['Translation'] was not set yet.
Testing $record['Translation'] first fixed it


vendethiel:
Seems to me like the isset should fail in that case.

alquerci:
This PR duplicate #184 

See https://github.com/FriendsOfSymfony1/symfony1/pull/184#issuecomment-406818771 for the technical reason.

PR #206, opened 2019-02-20
Fix autoloading for locales with different lowercase rules

sbeining:
The class map generation and the lookup should use a neutral locale to
generate the lowercase classnames.
Without this fix locales like turkish will not be able to load classes
with an uppercase „I“ for example.


j0k3r:
Looks an interesting fix, could you provide a test to cover that changes? Thanks.

PR #209, opened 2019-04-01
Allow 0 and '0' as false for a sfFormInputCheckbox and return null, if controller can't be found as promised in the docblock.

thirsch:


thirsch:
Before changing unit tests: What's your opinion, should '0' and 0 be a "valid" unchecked status?

alquerci:
@thirsch Thank you to spot this issue.

I recommends to take the same spec as Symfony4.

thirsch:
If I got it right, sf4 will allow the user to define the false-values: https://github.com/symfony/form/blob/189f2780738bfc0df71574c131fdb4b809dc2de8/Extension/Core/Type/CheckboxType.php, https://symfony.com/doc/current/reference/forms/types/checkbox.html#false-values - The default seems to be "null". What the sf4-checkbox really will do if false, 0, '' or '0' is provided maybe should be tested, but is a bit out of scope of my current project.

alquerci:
Thank you to dig on it.

> What the sf4-checkbox really will do if false, 0, '' or '0' is provided

https://github.com/symfony/form/blob/v4.2.5/Extension/Core/Type/CheckboxType.php#L45
https://github.com/symfony/form/blob/v4.2.5/Extension/Core/DataTransformer/BooleanToStringTransformer.php#L43-L63

Then the same as sf1 except that sf1 accept any value and not throw an exception.

---

To be sure to avoid any BC breaks. I suggest to add `false_values` option as Symfony do it.

---

Side note:
Please create a dedicated branch where you push this patch on it.
Then create a new PR with this branch.

thirsch:
I'll be back with another PR after adjusting it. But maybe not within the next couple of days.

PR #213, opened 2019-08-27
Allow redefinition of default error messages by a `sfValidator` subclass

hugochinchilla:


e1himself:
I agree with Alexandre. 

There is already a way to override the default messages from a child class.

```php
class UuidValidator extends sfValidatorBase 
{
    public function configure() {
        $this->setMessage('required', 'UUID is required.');
        $this->setMessage('invalid', 'Given UUID value is not valid.');
    }

    // ...
}

hugochinchilla:
You're right. We keep our own set of patches to symfony1 and I was trying to keep the differences with upstream as little as possible, this seemed simple enought to be accepted without discussion and tought that adding it to upstream would shorten the list of custom patches.

But you're right and we just need to override the configure method instead of the property definition.

PR #214, opened 2019-08-27
Add support for component argument resolving

hugochinchilla:
This adds support to resolve arguments using the DIC.

I's been implemented in a way that is extremly easy to swap the `sfServiceContainer` by other implementation, we use this with a Psr compilant DIC and the code looks like this:

```

### CODE REVIEW

File: lib/action/sfParameterResolver.class.php [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+      $type = $param->getClass();
+
+      // handle case where request parameter was not type hinted
+      if (null === $type && $i === 0) {
+        $parameters[] = $this->request;
+        continue;
+      }
+
+      if (null === $type) {
+        throw new \Exception("Aditional parameters must be type hinted");
```
vendethiel: typo: Additional

hugochinchilla: thanks, it's fixed now

hugochinchilla:
@alquerci I'm not sure what are you proposing about using symfony2+ DIC, I worked with what is available on this version, I use my own container but took the effort of making this compatible with the default DIC implementation but easy enougth to swap implementations with custom ones.

If you are sugesting that this project should adpot symfony2+ DIC I agree with you but this is outside the scope of this proposal, it would be easy to change it all on a future iteration.

I have provided unit tests to check the correct behavior of the parameter resolver and also some functional tests to make sure all the integration works as expected.

alquerci:
@hugochinchilla The concern is that implementing auto-wiring is more harder than just doing what you done.

IMHO the only issue here is a conceptual bug about the non ability to extends the controller call method. Adding this ability is fully on the migration path to Symfony2+.

So I suggest you to provide a way to customize the controller method call to add the ability to everyone to use there own choice. For example on [Symfony2][0] (which is compatible with PHP 5.3 like as symfony1) the resolution is based on request parameter.

The harder will be on the controller helper but it is still possible to do it.

[0]: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php

hugochinchilla:
@alquerci again I'm not sure what you are asking here. When you say I should add a way to customize the controller call method what are you asking for? If you mean a way to customize which method the controller should execute this is already supported. If you mean another thing please explain with more detail (some pseudocode will help).

Also the purpose of this PR is not to add autowiring, autowiring can be acomplished by using another DIC which supports it and I already provided example code on how to use another container to get even more funcionality from this PR.

alquerci:
An example of code on `sfActions` it can be :

```php
public function execute($request)
{
    // dispatch action
    $actionToRun = 'execute'.ucfirst($this->getActionName());

    // ....

   return $this->doCallController($request, $actionToRun);
}

protected function doCallController($request, $actionToRun)
{
    return $this->$actionToRun($request);
}
```

This way it just need to overwrite `doCallController()`  in order to customise the logic.

This way there is no need to add additional code as it fix the main issue.

PR #215, opened 2019-11-22
Add support for SameSite session cookies

mentalstring:
Starting from PHP 7.3 there's native support for [SameSite cookies](https://web.dev/samesite-cookies-explained/) ([RFC6265bis](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02)) which requires using a new session_get_cookie_params() parameter syntax.

### DIFF
--- a/test/unit/response/sfWebResponseTest.php
+++ b/test/unit/response/sfWebResponseTest.php
@@ -288,7 +288,7 @@ $t->is($response->clearJavascripts(), array(), '->clearJavascripts() removes all
 // ->setCookie() ->getCookies()
 $t->diag('->setCookie() ->getCookies()');
 $response->setCookie('foo', 'bar');
-$t->is($response->getCookies(), array('foo' => array('name' => 'foo', 'value' => 'bar', 'expire' => null, 'path' => '/', 'domain' => '', 'secure' => false, 'httpOnly' => false)), '->setCookie() adds a cookie for the response');
+$t->is($response->getCookies(), array('foo' => array('name' => 'foo', 'value' => 'bar', 'expire' => null, 'path' => '/', 'domain' => '', 'secure' => false, 'httpOnly' => false, 'samesite' => '')), '->setCookie() adds a cookie for the response');
 
 // ->setHeaderOnly() ->getHeaderOnly()
 $t->diag('->setHeaderOnly() ->isHeaderOnly()');
--- a/lib/response/sfWebResponse.class.php
+++ b/lib/response/sfWebResponse.class.php
@@ -158,10 +158,11 @@ class sfWebResponse extends sfResponse
    * @param  string  $domain    Domain name
    * @param  bool    $secure    If secure
    * @param  bool    $httpOnly  If uses only HTTP
+   * @param  string  $samesite  SameSite cookies
    *
    * @throws <b>sfException</b> If fails to set the cookie
    */
-  public function setCookie($name, $value, $expire = null, $path = '/', $domain = '', $secure = false, $httpOnly = false)
+  public function setCookie($name, $value, $expire = null, $path = '/', $domain = '', $secure = false, $httpOnly = false, $samesite = '')
   {
     if ($expire !== null)
     {
@@ -187,6 +188,7 @@ class sfWebResponse extends sfResponse
       'domain'   => $domain,
       'secure'   => $secure ? true : false,
       'httpOnly' => $httpOnly,
+      'samesite' => $samesite
     );
   }
 
@@ -365,7 +367,18 @@ class sfWebResponse extends sfResponse
     // cookies
     foreach ($this->cookies as $cookie)
     {
-      setrawcookie($cookie['name'], $cookie['value'], $cookie['expire'], $cookie['path'], $cookie['domain'], $cookie['secure'], $cookie['httpOnly']);
+      if (PHP_VERSION_ID < 70300) {
+        setrawcookie($cookie['name'], $cookie['value'], $cookie['expire'], $cookie['path'], $cookie['domain'], $cookie['secure'], $cookie['httpOnly']);
+      } else {
+        setrawcookie($cookie['name'], $cookie['value'], array(
+          'expires'  => $cookie['expire'],
+          'path'     => $cookie['path'],
+          'domain'   => $cookie['domain'],
+          'secure'   => $cookie['secure'],
+          'httpOnly' => $cookie['httpOnly'],
+          'samesite' => $cookie['samesite'],
+        ));
+      }
 
       if ($this->options['logging'])
       {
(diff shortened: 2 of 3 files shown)


michilehr:
Maybe add sfWebResponse->setCookie to be feature complete?

mentalstring:
Makes sense — should be good now.

test/unit/response/sfWebResponseTest.php
```diff
@@ -288,7 +288,7 @@ public function normalizeHeaderName($name)
 // ->setCookie() ->getCookies()
 $t->diag('->setCookie() ->getCookies()');
 $response->setCookie('foo', 'bar');
-$t->is($response->getCookies(), array('foo' => array('name' => 'foo', 'value' => 'bar', 'expire' => null, 'path' => '/', 'domain' => '', 'secure' => false, 'httpOnly' => false)), '->setCookie() adds a cookie for the response');
+$t->is($response->getCookies(), array('foo' => array('name' => 'foo', 'value' => 'bar', 'expire' => null, 'path' => '/', 'domain' => '', 'secure' => false, 'httpOnly' => false, 'samesite' => '')), '->setCookie() adds a cookie for the response');
```
alquerci: This modification tell that there is a BC break.

As the feature is only for `PHP >= 7.3` then current tests should not be touch.
But add a test to target specific version.

mkopinsky:
It looks like this PR was never merged. @alquerci does this just need a PHP-7.3-specific test in sfWebResponseTest.php?

test/unit/response/sfWebResponseTest.php
mkopinsky: This is not a BC break. The existence of `['samesite' => '']` in the output of getCookies() does not affect behavior - it gets ignored by `setrawcookie` and doesn't get sent in the header.

```php
<?php
// Falsy values are simply ignored - don't get sent in the header by PHP
setrawcookie('samesite_null', 'testing', [
  'samesite' => null
]);
setrawcookie('samesite_emptystring', 'testing', [
  'samesite' => ''
]);
setrawcookie('samesite_false', 'testing', [
  'samesite' => false
]);

// Other values, whether legit or not, do get sent
setrawcookie('samesite_gibberish', 'testing', [
  'samesite' => 'asqewrqewrq'
]);
setrawcookie('samesite_none', 'testing', [
  'samesite' => 'none'
]);
?>
Here's the body content
```

```
❯ curl -I http://localhost/cookie.php
HTTP/1.1 200 OK
Date: Thu, 05 Oct 2023 23:33:55 GMT
Server: Apache/2.4.57 (Unix) PHP/7.4.33
X-Powered-By: PHP/7.4.33
Set-Cookie: samesite_null=testing
Set-Cookie: samesite_emptystring=testing
Set-Cookie: samesite_false=testing
Set-Cookie: samesite_gibberish=testing; SameSite=asqewrqewrq
Set-Cookie: samesite_none=testing; SameSite=none
Content-Type: text/html; charset=UTF-8
```

alquerci: BC break does not impact PHP >= 7.3 but for PHP < 7.3.

What is the problem to have behaviour depending on PHP version?

mkopinsky: The behavior *does* depend on PHP version: https://github.com/FriendsOfSymfony1/symfony1/pull/215/files#diff-5aa2248855efa9ca1219f04cb7575c93a0a1f4af85e30ff32013bf84afcbf0d5R370-R381

In PHP < 7.3, it calls `session_set_cookie_params` in the original form. In PHP >= 7.3, it uses the new form (passing options as an array, including the `samesite` setting which is a no-op unless a value has been explicitly set). Yes, `$response->getCookies()` returns something different, but I don't think returning an extra key in an array from a getter of an internal property should be considered a BC break.

alquerci: What I think
---------------

As you said: The behaviour does depend on PHP version.
And tests are a proof that behaviour work as it should.

The tests should expose this behaviour.

`getCookies()` is parts of the public interface, not the private one.
So it should not lie on what value is passed to PHP internal functions.


In practice
-------------

Let's say that we can spy arguments passed to functions:
* `session_set_cookie_params`
* `setrawcookie`

On those tests, we will see something like that:

```php
// test/unit/**

if (PHP_VERSION_ID < 70300) {
    // behaviour without samesite
    givenCookie(array(
        'name' => 'some cookie',
        ...
    ));

    whenSendHttpHeaders();
    
    assertCookies(array(
        'cookie name' => array(
            'name' => 'some cookie',
            ...
        )
    ));
    assertSetrawcookieWasCalledWith('some cookie', ..., array(
        ...
    ));
} else {
    // behaviour with samesite
    givenCookie(array(
        'name' => 'some cookie',
        ...,
        'samesite' => 'samesite_value'
    ));

    whenSendHttpHeaders();
    
    assertCookies(array(
        'cookie name' => array(
            'name' => 'some cookie',
            ...
            'samesite' => 'samesite_value'
        )
    ));
    assertSetrawcookieWasCalledWith('some cookie', ..., array(
        ...,
        'samesite' => 'samesite_value'
    ));
}
```

thePanz:
@alquerci @mkopinsky (damn, that's quite an old ticket :sweat_smile: )

My plan is to move to PHP v7.4 as min version as a first step.
... and later to require PHP v8.1 as min version by removing all PHP v5.x and 7.x support

WDYT?

alquerci:
@thePanz I think that make sf1 compatible with latest PHP version is another topic than removing support for old PHP versions.

What problem will be fixed by removing all PHP v5.x and 7.x support?

My point of view
------------------

We should be able to have information about the usage of sf1 in terms of, at least:
* PHP version used
* sf1 version used

Because the segment of sf1 usage is the main target of the purpose of the support provided by FriendsOfSymfony1.

The main goal is to provide an easy path to migrate to the latest Symfony LTS version, or at least the first Symfony2 version.
"at least the first Symfony2 version" because Symfony2 itself have a migration path.

Nothing new as _FriendsOfSymfony1 organization what next?_ #201

_Provide a list of instructions how to switch from symfony 1.4.20 to FriendsOfSymfony1_ #244 
is the first step, the next one is from FriendsOfSymfony1 to Symfony2+

Do you agree with that goal?


### What I know

* Removing PHP <=7.2 version support will block the migration to Symfony 2.8.
    * Because Symfony 2.8 supports PHP `>=5.3.2 <=7.2`

| Symfony version | PHP version supported |
|--------|--------|
| 1.4 | 5.2 <= version <= 5.3 |
| 1.5.15 | 5.3 <= version <= 8.2 |
| 1.5.16-dev | 7.1 <= version <= 8.2 |
| 2.8 | 5.3.2 <= version <= 7.2 |
| 3.4 | 5.5.9 <= version <= 7.4 |
| 4.4 | 7.1.3 <= version <= 7.4 |
| 6.4 | 8.1 <= version <= 8.2 |

### Hypothesis

* Does migrate from sf1.4 to latest Symfony version is possible? Who did it?
* Does migrate from FriendsOfSymfony1 to Symfony2 is possible?

That's why, like a production management, we need to know users of sf1.

connorhu:
> Removing PHP <=7.2 version support will block the migration to Symfony 2.8

Who wants to migrate to an unsupported outdated version of symfony?

thePanz:
Replaced by #387

PR #216, opened 2019-11-22
Fix travis builds

mentalstring:
- running PHP 5.5 and 5.6 with Trusty makes it work again;
- removed unused travis key 'sudo';

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: php
 
 php:
-    - 5.5
-    - 5.6
     - 7.0
     - 7.1
     - 7.2
     - nightly
+
+dist: trusty
 
 # run build against nightly but allow them to fail
 matrix:
@@ -16,13 +16,12 @@ matrix:
     include:
         - php: 5.3
           dist: precise
-          sudo: required
         - php: 5.4
           dist: precise
-          sudo: required
-
-# faster builds on new travis setup not using sudo
-sudo: false
+        - php: 5.5
+          dist: trusty
+        - php: 5.6
+          dist: trusty
 
 services:
     - memcached
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ php:
     - 7.2
     - nightly
 
-dist: trusty
+dist: xenial
 
 # run build against nightly but allow them to fail
 matrix:


.travis.yml
```diff
@@ -8,21 +8,13 @@ php:
...
     fast_finish: true
     allow_failures:
         - php: nightly
-    include:
-        - php: 5.3
-          dist: precise
-          sudo: required
-        - php: 5.4
-          dist: precise
-          sudo: required
```
alquerci: 5.3 and 5.4 version are supported then please do not remove it.

mentalstring: I don't have much experience with Travis but I think the previous configuration wasn't testing 5.3 and 5.4 already?

alquerci: That's possible but let's run tests on all supported php version but at least on the first one.

alquerci: Hum, not https://travis-ci.org/FriendsOfSymfony1/symfony1/builds/577819519

Look at the end of the list.

mentalstring: You are right. As I said, I don't have much experience with Travis.

The issue seems to be that 5.3&5.4 needs to run with precise, but they are not picking up that configuration from .travis.yml. Any idea why?

alquerci: It is executed you can revert https://github.com/FriendsOfSymfony1/symfony1/pull/216/commits/f6a62b7e7c02c36a2499102df20a8824359b775a

And all will going good.

alquerci:
@mentalstring Can you take the opportunity to add PHP 7.3 and 7.4 on allow failure as currently nightly is PHP 8.0.0-dev <3

Thank you to take care of this.

mentalstring:
@alquerci No need to allow failure for 7.3 as it already passes all tests just fine. I'm planning on submitting a separate PR for adding PHP 7.3 to the list (and documentation)  — I was just trying to get Travis to pass first.

Now it seems that it run and pass on all PHP versions. Can you check?

.travis.yml
```diff
@@ -8,6 +8,8 @@ php:
     - 7.2
     - nightly
 
+dist: trusty
```
alquerci: [minor] Just for readability I suggest to move this line on the first line.

mentalstring:
Sorry, have made another update to the PR: better to specify the distro only for 5.5 and 5.6 that were the ones failing and leave the other versions run on what travis thinks is up to date — might be necessary to run newer/future versions of PHP.

alquerci:
@mentalstring Then let's make it explicit by using `dist: xenial` (on the top on the file), what do you think?

mentalstring:
What's the advantage? The way I see it, it's best not to lock it to a particular distro if we don't have to, as that way we can become aware of other tangent changes (eg: OS, libs, different distros) that can be relevant to run our tests against. We can always lock to a particular distro later if newer ones become incompatible to run these tests (as we are doing now with 5.5/5.6).

alquerci:
> What's the advantage?

The control on the infrastructure then to not have the same issue when boinc will be the default distribution.

We should focus on more important things.

mentalstring:
Anything missing to get this merged?

alquerci:
@mentalstring Nothing is missing to get this merged except the merger is available.

cc @thePanz

PR #218, opened 2019-11-27
Backport fix for cve-2019-18888

adrienperonnet:
`guessFromFileBinary` may also be impacted by [cve-2019-18888](https://symfony.com/blog/cve-2019-18888-prevent-argument-injection-in-a-mimetypeguesser) where
the provided file paths were not being properly escaped before being used.

This PR backports the fix from commit https://github.com/symfony/symfony/commit/691486e43ce0e4893cd703e221bafc10a871f365 to make sure the issue is not exploitable.

### DIFF
--- a/lib/validator/sfValidatorFile.class.php
+++ b/lib/validator/sfValidatorFile.class.php
@@ -263,7 +263,7 @@ class sfValidatorFile extends sfValidatorBase
   {
     ob_start();
     //need to use --mime instead of -i. see #6641
-    $cmd = 'file -b --mime %s -- 2>/dev/null';
+    $cmd = 'file -b --mime -- %s 2>/dev/null';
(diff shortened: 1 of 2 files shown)


lib/validator/sfValidatorFile.class.php
```diff
@@ -263,7 +263,9 @@ protected function guessFromFileBinary($file)
   {
     ob_start();
     //need to use --mime instead of -i. see #6641
-    passthru(sprintf('file -b --mime %s 2>/dev/null', escapeshellarg($file)), $return);
+    $cmd = 'file -b --mime %s -- 2>/dev/null';
```
alquerci: The path need to be placed after the double dash.

adrienperonnet: Well spotted thanks 👍

PR #224, opened 2020-03-22
Test PHP 7.3 and 7.4

W0rma:
* Test with PHP 7.3 and 7.4
* Fix array and string offset access using curly braces (deprecated since PHP 7.4 - see https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.array-string-access-curly-brace)
* Call [`get_magic_quotes_gpc()`](https://www.php.net/manual/en/function.get-magic-quotes-gpc.php) only for PHP < 5.4. It always returned false since PHP 5.4 and is deprecated since PHP 7.4
* Fix Travis config for PHP 5.5 and 5.6

### DIFF
--- a/lib/request/sfWebRequest.class.php
+++ b/lib/request/sfWebRequest.class.php
@@ -75,7 +75,14 @@ class sfWebRequest extends sfRequest
     parent::initialize($dispatcher, $parameters, $attributes, $options);
 
     // GET parameters
-    $this->getParameters = get_magic_quotes_gpc() ? sfToolkit::stripslashesDeep($_GET) : $_GET;
+    if (version_compare(PHP_VERSION, '5.4', '<') && get_magic_quotes_gpc())
+    {
+      $this->getParameters = sfToolkit::stripslashesDeep($_GET);
+    }
+    else
+    {
+      $this->getParameters = $_GET;
+    }
     $this->parameterHolder->add($this->getParameters);
 
     $postParameters = $_POST;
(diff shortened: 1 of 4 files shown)


lib/request/sfWebRequest.class.php
```diff
@@ -75,7 +75,14 @@ public function initialize(sfEventDispatcher $dispatcher, $parameters = array(),
     parent::initialize($dispatcher, $parameters, $attributes, $options);
 
     // GET parameters
-    $this->getParameters = get_magic_quotes_gpc() ? sfToolkit::stripslashesDeep($_GET) : $_GET;
+    if (version_compare(PHP_VERSION, '5.4', '<') && get_magic_quotes_gpc())
```
alquerci: I recommend to use `7.4.0-dev` instead of `7.4`.

>  Note that pre-release versions, such as 5.3.0-dev, are considered lower than their final release counterparts (like 5.3.0). 

source https://www.php.net/manual/en/function.version-compare.php

e1himself: @alquerci you mean `5.4.0-dev`, right? :)

alquerci: Ha, yes.

I make it as this patch is related to PHP 7.4 support. :)

W0rma: Thank you for the hint. Adjusted to `5.4.0-dev`

mentalstring:
I don't make a point in having my PR in — this one does the same and beyond 👍— I'll close #216 .

PR #227, opened 2020-04-15
Fix php 7.4 "Trying to access array offset on value of type null" notice on isCSRFProtected()

valepu:
I was having the "Trying to access array offset on value of type null" notice on this line in some pages of my website when using php 7.4. This should fix it

### DIFF
--- a/lib/form/sfForm.class.php
+++ b/lib/form/sfForm.class.php
@@ -944,7 +944,7 @@ class sfForm implements ArrayAccess, Iterator, Countable
    */
   public function isCSRFProtected()
   {
-    return null !== $this->validatorSchema[self::$CSRFFieldName];
+    return isset($this->validatorSchema) && null !== $this->validatorSchema[self::$CSRFFieldName];
   }
 
   /**


alquerci:
Hello @valepu,

Thank you to raise this issue.

As I see the property is set on the constructor and I think that it is not intended to be set to null after.
https://github.com/FriendsOfSymfony1/symfony1/blob/e06b85696a61659da71b47b6a9d26df3c6dd11cb/lib/form/sfForm.class.php#L70

Do you have the stack trace in order to have more infos ?

So the proper fi will be to avoid to be setted to `null`.

Is it possible to add a tests for this in order to point the exact case when the bug occured.

valepu:
Hi,
At the moment i don't have any stack trace since i had it only as a notice. I have looked in the whole project (including symfony directory) and i couldn't find a line where validatorschema is set to null except when the parameter is initialized. 
Yet when adding that check the notice disappeared so maybe the issue could be more subtle. I'll try to debug the code and see what is really happening.

alquerci:
@valepu 

> used on method $this->setDefaults()

And `$this->setDefaults()` is called on the constructor before `validatorSchema` initialization.

valepu:
Ah you are right! Didn't even notice it. Would it be more correct to move setDefaults after, then?

alquerci:
There is on issue here about to test that as this case is already covered but not reported as error reporting on test ignore at least `E_NOTICE`.

PR #228, opened 2020-05-02
Php 7.4 Incompatibility Fixes

pathumhdes:
1) Passing the $glue and $pieces parameters in reverse order to implode has been deprecated since PHP 7.4; $glue should be the first parameter and $pieces the second
2) Curly brace syntax for accessing array elements and string offsets has been deprecated in PHP 7.4

### DIFF
--- a/lib/i18n/sfDateFormat.class.php
+++ b/lib/i18n/sfDateFormat.class.php
@@ -229,7 +229,7 @@ class sfDateFormat
     for ($i = 0, $max = count($tokens); $i < $max; $i++)
     {
       $pattern = $tokens[$i];
-      if ($pattern[0] == "'" && $pattern[(strlen($pattern) - 1)] == "'")
+      if ($pattern[0] == "'" && $pattern[strlen($pattern) - 1] == "'")
       {
         $tokens[$i] = str_replace('``````', '\'', preg_replace('/(^\')|(\'$)/', '', $pattern));
       }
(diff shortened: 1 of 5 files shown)


mkopinsky:
It looks like travis tests failed because of a network fluke. Could one of you retrigger the tests?

spirit-q2:
@pathumhdes just squash two of your three commits and it will trigger re-run the tests

pathumhdes:
@spirit-q2 @mkopinsky The tests are passing now.

lib/i18n/sfDateFormat.class.php
```diff
@@ -229,7 +229,7 @@ public function format($time, $pattern = 'F', $inputPattern = null, $charset = '
     for ($i = 0, $max = count($tokens); $i < $max; $i++)
     {
       $pattern = $tokens[$i];
-      if ($pattern[0] == "'" && $pattern{strlen($pattern) - 1} == "'")
+      if ($pattern[0] == "'" && $pattern[(strlen($pattern) - 1)] == "'")
```
mentalstring, line 232: Very minor thing, but could the inner parentheses be skipped here? They are not needed and don't improve readability.

lib/i18n/sfNumberFormat.class.php
```diff
@@ -193,7 +193,7 @@ protected function formatInteger($string)
       // now for the integer groupings
       for ($i = 0; $i < $len; $i++)
       {
-        $char = $string{$len - $i - 1};
+        $char = $string[($len - $i - 1)];
```
mentalstring, line 196: Same here.

pathumhdes: done

j0k3r:
I was wondering why is there nothing about these _Incompatibilities_ reported in Travis log?
https://travis-ci.org/github/FriendsOfSymfony1/symfony1/jobs/670507708

I there no test to cover each case? If so, would you mind adding some?

Issue #183, opened 2018-02-10
Warning: ini_set(): A session is active. You cannot change the session module's ini settings at this time in symfony1\lib\user\sfBasicSecurityUser.class.php on line 257


sdkcarlos:
I've upgraded Symfony 1 to this version, however i'm getting this exception in PHP 7.2:

> Warning: ini_set(): A session is active. You cannot change the session module's ini settings at this time in  project\vendor\lexpress\symfony1\lib\user\sfBasicSecurityUser.class.php on line 257

I am not sure if this bug is related to the fixes to make Sf compatible with PHP 7 or is an error of the library.

e1himself:
It seems like a bug, visible only with specific conditions.

https://github.com/LExpress/symfony1/blob/master/lib/user/sfBasicSecurityUser.class.php#L255-L258

Try either lowering user session `timeout` in your project *factories.yml* (see example here: https://github.com/LExpress/symfony1/blob/master/lib/config/config/factories.yml#L32)

or `session.gc_maxlifetime` in your *php.ini*

or submit a Pull Request to fix it properly :)

gileri:
Raising session.gc_maxlifetime is indeed the solution, but I don't think it is doable in `sfBasicSecurityUser`.

It must be done before `session_start()`, and `sfSessionStorage` doesn't seem coupled with `sfBasicSecurityUser` so it can't raise `session.gc_maxlifetime` in respect of what `sfBasicSecurityUser` it may store before calling `session_start()`.

I think the only way to solve this is to make https://github.com/LExpress/symfony1/blob/master/lib/user/sfBasicSecurityUser.class.php#L255-L258 into a warning/error instead of trying to raise `session.gc_maxlifetime`.

pbowyer:
With PHP 7.2 (but not listed in the PHP changelog) ini modifications were made to fail - see https://gist.github.com/tpunt/d9fd57c87c37a06a1057e970a90cd8df.

My fix is:
```
    if (session_status() !== PHP_SESSION_ACTIVE && ini_get('session.gc_maxlifetime') < $this->options['timeout'])
    {
      ini_set('session.gc_maxlifetime', $this->options['timeout']);
    }
```

spirit-q2:
Guys, check PR above :point_up:

PR #229, opened 2020-05-06
Fix for `Warning: ini_set(): A session is active...`

spirit-q2:
Addresses the issue: https://github.com/FriendsOfSymfony1/symfony1/issues/183

Changing session settings shouldn't be done on `sfBasicSecurityUser` class. Instead it should be done near (and **_before_**) the call to `session_start()` in `sfSessionStorage`

### DIFF
--- a/lib/config/sfFactoryConfigHandler.class.php
+++ b/lib/config/sfFactoryConfigHandler.class.php
@@ -109,6 +109,10 @@ class sfFactoryConfigHandler extends sfYamlConfigHandler
             unset($parameters['database']);
           }
 
+          if (isset($config['user']['param']['timeout'])) {
+            $defaultParameters[] = sprintf("'gc_maxlifetime' => %d,", $config['user']['param']['timeout']);
+          }
+
           $instances[] = sprintf("  \$class = sfConfig::get('sf_factory_storage', '%s');\n  \$this->factories['storage'] = new \$class(array_merge(array(\n%s\n), sfConfig::get('sf_factory_storage_parameters', %s)));", $class, implode("\n", $defaultParameters), var_export($parameters, true));
           break;
 
--- a/lib/storage/sfSessionStorage.class.php
+++ b/lib/storage/sfSessionStorage.class.php
@@ -63,6 +63,7 @@ class sfSessionStorage extends sfStorage
       'session_cookie_secure'   => $cookieDefaults['secure'],
       'session_cookie_httponly' => isset($cookieDefaults['httponly']) ? $cookieDefaults['httponly'] : false,
       'session_cache_limiter'   => null,
+      'gc_maxlifetime'          => 1800,
     ), $options);
 
     // initialize parent
@@ -88,6 +89,12 @@ class sfSessionStorage extends sfStorage
     if (null !== $this->options['session_cache_limiter'])
     {
       session_cache_limiter($this->options['session_cache_limiter']);
+    }
+
+    // force the max lifetime for session garbage collector to be greater than timeout
+    if (ini_get('session.gc_maxlifetime') < $this->options['gc_maxlifetime'])
+    {
+      ini_set('session.gc_maxlifetime', $this->options['gc_maxlifetime']);
     }
 
     if ($this->options['auto_start'] && !self::$sessionStarted)
--- a/lib/user/sfBasicSecurityUser.class.php
+++ b/lib/user/sfBasicSecurityUser.class.php
@@ -4,7 +4,7 @@
  * This file is part of the symfony package.
  * (c) 2004-2006 Fabien Potencier <fabien.potencier@symfony-project.com>
  * (c) 2004-2006 Sean Kerr <sean@code-box.org>
- * 
+ *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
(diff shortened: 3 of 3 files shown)


spirit-q2:
https://travis-ci.org/github/FriendsOfSymfony1/symfony1/jobs/683856543 doesn't look like a related error :thinking: 

```
  [Composer\Downloader\TransportException]  
  Peer fingerprint did not match

The command "composer install" failed and exited with 1 during .
```

spirit-q2:
@alquerci hello, have some time to take a look? ;)

lib/user/sfBasicSecurityUser.class.php
```diff
@@ -251,12 +251,6 @@ public function initialize(sfEventDispatcher $dispatcher, sfStorage $storage, $o
       $this->options['timeout'] = 1800;
     }
```
alquerci, line 252: This logic need also to be copied before access of `$this->options['timeout']` on `sfSessionStorage`.

-> https://github.com/FriendsOfSymfony1/symfony1/blob/0f6bdd8814e33a3713fc36df780dfe583d8033b1/lib/storage/sfSessionStorage.class.php#L56-L66

spirit-q2: thanks @alquerci  for your review :)

> Need to check deeper but on a first approximation there can be a BC break regarding session timeout.

I wonder how's that? Current code on `sfBasicSecurityUser` is clearly no-op. It just don't send a warning, but it's not working for sure.

> Also need to propagate this new factory parameter to all other session classes

no needs, as all session classes derives from `sfSessionStorage` and calls to `parent::initialize(...)`

>where both configuration needs to be in synchronization

A `user` can get the option from a `storage` and if it exists use it (check updated PR).

Regarding the name of parameter, I'd rather leave it as `timeout` as it's more meaningful when we are reading the config.

alquerci: > A user can get the option from a storage and if it exists use it

It is exactly the inverse to do as the current timeout is given on user scope.

This behaviour is not going to change.

---

### Actions

1. Keep the same configuration schema.
2. To be able to fetch the timeout configuration from `user` factory configuration.
3. Before `session_start()` if `timeout` is greater than `gc_maxlifetime` then `gc_maxlifetime = timeout`

### Issues to solve

* `user` depends on `storage` but `storage` depends on `user` config
  - Track to patch -> on the `sfFactoryConfigHandler`

spirit-q2: you killing me, as always :stuck_out_tongue: alright let me check.

spirit-q2: btw, @alquerci, can you advise me how to run tests on the library properly? :pray:

spirit-q2:
@alquerci please, check latest version ;)

lib/user/sfBasicSecurityUser.class.php
alquerci: @spirit-q2 Look at how tests are executed on travis configuration.

Using official docker php:7.2-cli image with following extensions

* pdo
* pdo_mysql                                                                                                                                              
* mbstring

And link it to a MySql database with an empty database named `test`.
The official mysql:5.5 docker image can make the stuff with env `MYSQL_ALLOW_EMPTY_PASSWORD=yes`.

lib/storage/sfSessionStorage.class.php
```diff
@@ -68,6 +68,11 @@ public function initialize($options = null)
     // initialize parent
     parent::initialize($options);
 
+    if (!array_key_exists('gc_maxlifetime', $this->options))
+    {
+      $this->options['gc_maxlifetime'] = 1800;
+    }
```
alquerci, line 74: This configuration check can be merged on default parameters values used on the first parameter of `array_merge()`, see on line 56.

lib/config/sfFactoryConfigHandler.class.php
```diff
@@ -109,6 +109,12 @@ public function execute($configFiles)
             unset($parameters['database']);
           }
 
+          $sessionGcMaxlifetime = 1800;
+          if (isset($config['user']['timeout'])) {
```
alquerci, line 113: The `timeout` parameter is located on `$config['user']['param']['timeout']`.

lib/config/sfFactoryConfigHandler.class.php
```diff
@@ -109,6 +109,12 @@ public function execute($configFiles)
...
+            $sessionGcMaxlifetime = $config['user']['timeout'];
+          }
+          $defaultParameters[] = sprintf("'gc_maxlifetime' => %d,", $sessionGcMaxlifetime);
```
alquerci, line 116: Only append this new configuration if it is set on user factory configuration.

Let keep the default configuration value on `sfSessionStorage` class to avoid duplication and confusion.

spirit-q2:
@alquerci updated all comments

spirit-q2:
@j0k3r @jeromemacias @e1himself @GromNaN Hello guys! Can anyone check/approve/merge?

ncharlot:
Interested by this fix too.

spirit-q2:
@j0k3r @jeromemacias @e1himself @GromNaN 

Guys, any chances to get this merged?
and these PRs as well: https://github.com/FriendsOfSymfony1/symfony1/pull/230, https://github.com/FriendsOfSymfony1/symfony1/pull/228

It's the only blocker to move whole project to `php7.4` for me (and I guess for many other people who are willing to move to new `php`)

alquerci:
cc @thePanz :point_up:

j0k3r:
I’ll take care of them

Issue #234, opened 2020-06-19
sfValidatorBoolean - clean method returns true if the tainted value is 0 (integer zero)


pathumhdes:
The clean method in sfValidatorBoolen returns false if the tainted value is '0' (string zero) which is correct. But it returns true if the tainted value is 0 (integer zero)

PR #235, opened 2020-06-19
Fixing Issue #234

pathumhdes:
Fixing Issue #234 - sfValidatorBoolean - clean method returns true if the tainted value is 0 (integer zero)


thePanz (approved):

thePanz:
Thanks @pathumhdes , merged!

PR #236, opened 2020-06-20
[Lime] Fix backward compatibility break for handle_exception signature

alquerci:
Throwables were introduced since PHP 7.
http://php.net/manual/en/class.throwable.php

Now `lime_test::handle_exception()` method works properly with any `Exception` and `Error` instances.

This fix the backward compatibility breaking changes introduce on:

5ea8d656 exception_handler signature changed in php7


alquerci:
Closed as fixed on #266

Issue #161, opened 2017-06-26
Update swiftmailer to >= 5.4.4


KatharinaSt:
Current version is 6.0.1 so it would be great to be able to use the newest version.

As Swiftmailer 5.4.4 has a security vulnerability at least a minor update would be reasonable.
http://www.cvedetails.com/cve/CVE-2016-10074/

KatharinaSt:
This is the code I used to make sfMailer work with Swiftmailer 6.0.1:

Instantiation with $subject via "new":

    public function compose($from = null, $to = null, $subject = null, $body = null)
    {
        return (new Swift_Message($subject))
            ->setFrom($from)
            ->setTo($to)
            ->setBody($body)
        ;
    }

and type hint in send():

    public function send(Swift_Mime_SimpleMessage $message, &$failedRecipients = null)
    {
        //...
    }

j0k3r:
Feel free to provide a PR to update the composer.json, the submodule and sfMailer (be careful about compatibility)

rsantellan:
Migrate to swiftmailer 6 will break the php mail transport. That's ok?

j0k3r:
I'm not ok.
Maybe we can provide a different sfMailer or a way to configure it to be used with switf 5 or 6.

I've checked swiftmailer 6.

mentalstring:
Something to take into account is that Swiftmailer 6 has set PHP 7.0 as minimum.

j0k3r:
Which means we won't upgrade the version within symfony1 to avoid losing PHP 5 compatibility

mentalstring:
@j0k3r How about cloning sfMailer into a sfMailer6 of sorts with the needed changes to give the option of using swiftmailer 6 for those that are using PHP 7.0+? There would be a bit of code duplication, but on its own class it shouldn't interfere with the rest. Happy to try to put a PR together for this if you think this could be a viable option.

alquerci:
@mentalstring see #240

thirsch:
@KatharinaSt #240 has been merged. You can now use swiftmailer 6 :)

PR #237, opened 2020-09-07
Feature/swiftmailer 6 compat

thirsch:
We have added support for swiftmailer 6 by providing a second version of sfMailer that is compatible with swiftmailer 6. As we are installing symfony via composer, it's no issue for us to not using the submodule version. But is that ok for the upstream too?

Updating a previous symfony1 installation might require to change any static access from `sfMailer::` to `sfMailerBase::`. But before adjusting the doc etc. I wanted to check with you if this branch is of interest for you.

Solves #161


mentalstring:
FWIW,

I'd welcome support for Swiftmailer 6, specially if this also maintains support for v5.

I would be fine with not using the submodule approach, if that's the only roadblock.

thirsch:
I came to another, way easier approach without breaking any backward compatibility within the code. Maybe we can also keep the submodule and if the user is using the submodule, it will simply install ~5.2 and using the composer it will install ~5.2 or ~6.0 depending on the available php version. Let me do some tests for the changes and the submodule idea.

thirsch:
Yes, it's working! With the current change we are backward compatible and we even can keep the submodule approach!

To have the tests working again, we would need #236 first. After that, we can check the status of the mailer tests.

lib/mailer/sfMailer.class.php
```diff
@@ -274,31 +334,6 @@ public function sendNextImmediately()
...
 
-  /**
-   * Sends the given message.
-   *
-   * @param Swift_Transport $transport         A transport instance
-   * @param string[]        &$failedRecipients An array of failures by-reference
-   *
-   * @return int|false The number of sent emails
-   */
-  public function send(Swift_Mime_Message $message, &$failedRecipients = null)
```
alquerci, line 285: Remove the type hint.

lib/mailer/sfMailer.class.php
```diff
@@ -227,26 +307,6 @@ public function setDeliveryAddress($address)
...
-   * @param string|array $from    The from address
-   * @param string|array $to      The recipient(s)
-   * @param string       $subject The subject
-   * @param string       $body    The body
-   *
-   * @return Swift_Message A Swift_Message instance
-   */
-  public function compose($from = null, $to = null, $subject = null, $body = null)
-  {
-    return Swift_Message::newInstance()
```
alquerci, line 242: replace with `new Swift_Message()`.

composer.json
```diff
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php" : ">=5.3.0",
-        "swiftmailer/swiftmailer": "~5.2"
+        "swiftmailer/swiftmailer": "~5.2|~6.0"
```
alquerci, line 8: The dependency constraint should be more strict like that => `^6.0`.

alquerci (changes requested):
Hello @thirsch ,

Thank you to propose this compatibility.

Look at line comments where I expose the minimum change to add SwiftMailer version 6 compatibility.

No need to introduce `sfMailerBase`.

Also the submodule should be left untouched.

lib/mailer/sfMailer.class.php
thirsch: Thanks, that was way to simple... ;-) I wasn't aware, that you can remove the typehint and be less specific while extending from another class.

composer.json
thirsch: Thanks, could you please explain the difference to me? From what I understood on https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators, it will be the same and only differ, if I would introduce the third digit.

thirsch:
> Hello @thirsch ,
> 
> Thank you to propose this compatibility.
> 
> Look at line comments where I expose the minimum change to add SwiftMailer version 6 compatibility.
> 
> No need to introduce `sfMailerBase`.
> 
> Also the submodule should be left untouched.

Thanks @alquerci, I'll close this pr and open a new one with only the relevant changes and without removing the submodule.

lib/mailer/sfMailer.class.php
alquerci: You rise me a doubt. But for sure the test have always right, and the result may depend on the PHP version.

composer.json
alquerci: The `~` include `7.0@dev` but not with `^`.

lib/mailer/sfMailer.class.php
thirsch: Yes it's working. On a new codebase I would not do prefer it that way, but on the legacy stuff it's sufficient to get the compat for swiftmailer 6 :)

alquerci: Yes, however, even on new codebase as the type constraint is respected with the call on the parent and the transport.

There is a goal to avoid using language version specific code as much as possible to avoid useless complexity.

But that's another story depending of the codebase. I give you me point of view.

composer.json
e1himself: > The `~` include `7.0@dev` but not with `^`.

Hm... Really? 

I think they explicitly mention that it will not install versions having major number different from `6` @alquerci 

> Note: Although `2.0-beta.1` is strictly before `2.0`, a version constraint like `~1.2` would not install it. As said above `~1.2` only means the `.2` can change but the `1.` part is fixed.

https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

alquerci: Hey, indeed I wasn't understand in details the difference.

Now yes.

> The ^ operator behaves very similarly, but it sticks closer to semantic versioning, and will always allow non-breaking updates.

https://getcomposer.org/doc/articles/versions.md#caret-version-range-

#learningAllDays

PR #239, opened 2020-09-08
Feature/display errors to stderr

thirsch:
This pr allows to set stderr as the target for display_errors. For us it is helpful during development, especially while upgrading to newer versions of PHP. The default (stdout) is equal to "On" which is deprecated since 5.2.4: https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors

### DIFF
--- a/lib/config/sfApplicationConfiguration.class.php
+++ b/lib/config/sfApplicationConfiguration.class.php
@@ -143,7 +143,8 @@ abstract class sfApplicationConfiguration extends ProjectConfiguration
     }
 
     // error settings
-    ini_set('display_errors', $this->isDebug() ? 'on' : 'off');
+    // PHP >= 5.2.4 supports stdout/stderr instead of bool here: https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors
+    ini_set('display_errors', $this->isDebug() ? sfConfig::get('sf_display_errors', 'stdout') : '');
     error_reporting(sfConfig::get('sf_error_reporting'));
 
     // initialize plugin configuration objects
--- a/WHATS_NEW.md
+++ b/WHATS_NEW.md
@@ -165,7 +165,8 @@ This allow you to remove all stylesheets and/or all javascripts added into the g
 Configuration
 -------------
 
-A new configuration `sf_upload_dir_name` contains 'uploads' has been added.
+* A new configuration `sf_upload_dir_name` contains 'uploads' has been added.
+* A new settings `sf_display_errors` has been added to change display_errors from stdout to stderr for example.
 
 Performance
 -----------


lib/config/sfApplicationConfiguration.class.php
```diff
@@ -143,7 +143,8 @@ public function initConfiguration()
     }
 
     // error settings
-    ini_set('display_errors', $this->isDebug() ? 'on' : 'off');
+    // PHP >= 5.2.4 supports stdout/stderr instead of bool here: https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors
+    ini_set('display_errors', $this->isDebug() ? sfConfig::get('sf_display_errors', 'stdout') : '');
```
alquerci, line 147: To fully tackle this part: the correct behaviour is found by newer Symfony version (2+) and become this:

```diff
-        ini_set('display_errors', $this->isDebug() ? 'on' : 'off');
+        if (null === $displayErrors = sfConfig::get('sf_display_errors')) {
+            $displayErrors = true; // Even when debug is off we want to see errors on output/console.
+        }
+        if ('cli' !== PHP_SAPI) {
+            ini_set('display_errors', 0);
+        } elseif ($displayErrors && (!ini_get('log_errors') || ini_get('error_log'))) {
+            // CLI - display errors only if they're not already logged to STDERR
+            ini_set('display_errors', 1);
+        }
```

This patch done this:

> Except for CLI - display errors only if they're not already logged to STDERR

See https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/ErrorHandler/Debug.php#L25-L30

alquerci (changes requested):
Hello @thirsch ,

Thanks for take time to tackle this issue.

Also need to add default configuration to `~` on
https://github.com/FriendsOfSymfony1/symfony1/blob/70db22f0ba9b8e13583ca90f13da93eac65729fd/lib/config/config/settings.yml#L36

lib/config/sfApplicationConfiguration.class.php
thirsch: Thanks, that would mean, display_errors is always off, if not run in cli. Should we also consider phpdbg as shown in the diff you referred to?

thirsch: And maybe we should execute the whole thing only if $this->isDebug() is on, right?

alquerci: Web context
-------

display_errors should always be off.
This cause issue on HTML, json, ... format rendering.

Console
-----

Like said the comment:
> display errors only if they're not already logged to STDERR

Problematic
--------

Fatal error are not handled by sf1 error handler, there are 2 solutions:

1. Use symfony/debug and disable error handling by sfWebDebugLogger
2. enable PHP ini `log_errors` to log the error on server logs.

---

What problematic do you try to solve? can you add more details on PR description?

---

For sure le latest version of the logic is on symfony/symfony repository.
The only effort is to make it compatible with PHP 5.3 =.> PHP 7.4

PR #240, opened 2020-09-09
Compat for Swiftmailer 6 but preserve compatibility with Swiftmailer 5.

thirsch:
This pr is the follow up for #237. Thanks for your review.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php" : ">=5.3.0",
-        "swiftmailer/swiftmailer": "~5.2"
+        "swiftmailer/swiftmailer": "~5.2|^6.0"
     },
     "require-dev": {
         "psr/log": "*"
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php" : ">=5.3.0",
-        "swiftmailer/swiftmailer": "~5.2|^6.0"
+        "swiftmailer/swiftmailer": "~5.2 || ^6.0"
     },
     "require-dev": {
         "psr/log": "*"
--- a/lib/mailer/sfMailer.class.php
+++ b/lib/mailer/sfMailer.class.php
@@ -239,7 +239,8 @@ class sfMailer extends Swift_Mailer
    */
   public function compose($from = null, $to = null, $subject = null, $body = null)
   {
-    return Swift_Message::newInstance()
+    $msg = new Swift_Message();
+    return $msg
       ->setFrom($from)
       ->setTo($to)
       ->setSubject($subject)
@@ -282,7 +283,7 @@ class sfMailer extends Swift_Mailer
    *
    * @return int|false The number of sent emails
    */
-  public function send(Swift_Mime_Message $message, &$failedRecipients = null)
+  public function send($message, &$failedRecipients = null)
   {
     if ($this->force)
     {
--- a/lib/mailer/sfMailer.class.php
+++ b/lib/mailer/sfMailer.class.php
@@ -240,6 +240,7 @@ class sfMailer extends Swift_Mailer
   public function compose($from = null, $to = null, $subject = null, $body = null)
   {
     $msg = new Swift_Message();
+
     return $msg
       ->setFrom($from)
       ->setTo($to)
@@ -278,8 +279,8 @@ class sfMailer extends Swift_Mailer
   /**
    * Sends the given message.
    *
-   * @param Swift_Transport $transport         A transport instance
-   * @param string[]        &$failedRecipients An array of failures by-reference
+   * @param Swift_Mime_Message|Swift_Mime_SimpleMessage $message           The message to send.
+   * @param string[]                                    &$failedRecipients An array of failures by-reference
    *
    * @return int|false The number of sent emails
    */
--- a/lib/mailer/sfNoMailer.class.php
+++ b/lib/mailer/sfNoMailer.class.php
@@ -56,7 +56,7 @@ class sfNoMailer
     return null;
   }
 
-  public function send(Swift_Mime_Message $message, &$failedRecipients = null)
+  public function send($message, &$failedRecipients = null)
   {
     return null;
   }


lib/mailer/sfMailer.class.php
```diff
@@ -239,7 +239,8 @@ public function setDeliveryAddress($address)
    */
   public function compose($from = null, $to = null, $subject = null, $body = null)
   {
-    return Swift_Message::newInstance()
+    $msg = new Swift_Message();
+    return $msg
```
alquerci, line 243: [minor] Can you add a new line before the `return` (for readability) ?

lib/mailer/sfMailer.class.php
```diff
@@ -282,7 +283,7 @@ public function sendNextImmediately()
    *
    * @return int|false The number of sent emails
    */
-  public function send(Swift_Mime_Message $message, &$failedRecipients = null)
+  public function send($message, &$failedRecipients = null)
```
alquerci, line 286: [minor] It will be a good opportunity to fix the PHPDoc of the first argument as it is not a `Swift_Transport` but a message. So document what type to pass depending on Swift Mailer version.

alquerci (changes requested):
LGTM but let's not forgot to add tests.

For that I suggest to add one job with SwiftMailer version 6 on travis build.

Do you know enough about travis-ci to do that change or you need help?

PS: you can update the same PR by just push another commit to `feature/swiftmailer-6-compat-2`.

lib/mailer/sfMailer.class.php
thirsch: I've added both allowed classes as `Swift_Mime_Message|Swift_Mime_SimpleMessage` in the docblock. Is that ok or would you prefer a typeless message and having the correct type explained in the description?

thirsch:
> LGTM but let's not forgot to add tests.
> 
> For that I suggest to add one job with SwiftMailer version 6 on travis build.
> 
> Do you know enough about travis-ci to do that change or you need help?
> 
> PS: you can update the same PR by just push another commit to `feature/swiftmailer-6-compat-2`.

Thanks for your re-review. I've not done a project on travis so far and help would be much appreciated. At the moment we are installing swiftmailer 5 if the tests are running on PHP < 7 and if we are running on PHP >= 7, it is already installing swiftmailer 6. Any specific point to test or would that be sufficient?

lib/mailer/sfMailer.class.php
alquerci: That's the minimum things to do in term of documentation as on each cases it works.

`Swift_Mime_Message` is not available on SwiftMailer version 6 then cannot be instantiate.

alquerci:
> Any specific point to test or would that be sufficient?

That's sufficient.

To execute the current test for each swift version
https://github.com/FriendsOfSymfony1/symfony1/blob/6185ffc7f31bfe68d37690b4d3533a5ab69279dd/test/unit/mailer/sfMailerTest.php

> help would be much appreciated

I will try to tackle this this week-end.

A trade-off here is to install swift version 6 on before install then revert the commit when travis build is done.

thirsch:
> > help would be much appreciated
> 
> I will try to tackle this this week-end.

Thanks

> A trade-off here is to install swift version 6 on before install then revert the commit when travis build is done.
> The next step will be to add a dedicated job.

At the moment `composer install` during the build process already installs swift version 5 on PHP < 7 and swift version 6 on PHP >= 7. Or do you plan do somehow force swift version 5 on PHP > 7 as an additional test for version 6?

alquerci:
> At the moment composer install during the build process already installs swift version 5 on PHP < 7 and swift version 6 on PHP >= 7. Or do you plan do somehow force swift version 5 on PHP > 7 as an additional test for version 6?

Ha, got it now (refreshing my memory => because composer install the latest possible version).

In that case I think that's enough except there is a bug on lowest versions.
Using `composer update --prefer-lowest` => :heavy_check_mark:

alquerci (approved):

mentalstring (approved):
LGTM 👍

lib/mailer/sfMailer.class.php
```diff
@@ -239,7 +239,9 @@ public function setDeliveryAddress($address)
    */
   public function compose($from = null, $to = null, $subject = null, $body = null)
   {
-    return Swift_Message::newInstance()
+    $msg = new Swift_Message();
```
thePanz, line 242: $msg is not needed :)
Why don't we `return new Swift_Message();` ?

composer.json
```diff
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php" : ">=5.3.0",
-        "swiftmailer/swiftmailer": "~5.2"
+        "swiftmailer/swiftmailer": "~5.2|^6.0"
```
thePanz, line 8: Can we use the recommended operator "||" here? (personal taste :) )
`~5.2 || ^6.0`

thePanz (changes requested):

lib/mailer/sfMailer.class.php
thirsch: Well, because there are a few method calls on the object. That would mean to `return (new Swift_Message())->setFrom($from)...;`. Should I change it? Don't think it's mission critical, but would be ok for me.

composer.json
thirsch: sure :)

thePanz (approved):

thePanz:
Thank you @thirsch , merged

thePanz:
Found out that `lib/plugins/sfDoctrinePlugin/lib/mailer/Swift_DoctrineSpool.class.php` class still references to `Swift_Mime_Message`, which is not around if the application installs SwiftMailer v6.

Any hints @thirsch ?

thirsch:
Good catch, @thePanz. We could do the same as in this pr by removing the type from the method signature and add both types to the phpdoc. I can create a pr for it, if you like.

thePanz:
> Good catch, @thePanz. We could do the same as in this pr by removing the type from the method signature and add both types to the phpdoc. I can create a pr for it, if you like.

Yes, please go ahead :+1: 
the tricky part: the interface is from the SwiftMailer itself, IIRC :)

thirsch:
> Yes, please go ahead 👍 the tricky part: the interface is from the SwiftMailer itself, IIRC :)

Ok, got it. We might be able to fix it with the solution I had before just removing the type. The pr is still available, so please have a look at it: https://github.com/FriendsOfSymfony1/symfony1/pull/237/files#diff-ab78635a7285fe1eca9b5c7fad605d21d3f09ccd2737c0af6bbf7022f66f95a7

As I'm on vacation next week, it will take a few days to come up with a proposal.

PR #242, opened 2020-12-17
Fix PHP 7.4 deprecation warnings

terdia:
```
public static function isPathAbsolute($path)
  {
    if ($path{0} === '/' || $path{0} === '\\' ||
        (strlen($path) > 3 && ctype_alpha($path{0}) &&
         $path{1} === ':' &&
         ($path{2} === '\\' || $path{2} === '/')
        )
       )
    {
      return true;
    }

    return false;
  }
```
Causes deprecation warning:

Deprecated: Array and string offset access syntax with curly braces is deprecated in `/var/www/vendor/lexpress/symfony1/lib/util/sfFinder.class.php on line 586`

`$this->getParameters = get_magic_quotes_gpc() ? sfToolkit::stripslashesDeep($_GET) : $_GET;`

Deprecated: Function get_magic_quotes_gpc() is deprecated in `/var/www/vendor/lexpress/symfony1/lib/request/sfWebRequest.class.php on line 78`

### DIFF
--- a/lib/request/sfWebRequest.class.php
+++ b/lib/request/sfWebRequest.class.php
@@ -75,7 +75,7 @@ class sfWebRequest extends sfRequest
     parent::initialize($dispatcher, $parameters, $attributes, $options);
 
     // GET parameters
-    $this->getParameters = get_magic_quotes_gpc() ? sfToolkit::stripslashesDeep($_GET) : $_GET;
+    $this->getParameters = $_GET;
     $this->parameterHolder->add($this->getParameters);
 
     $postParameters = $_POST;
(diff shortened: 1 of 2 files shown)


alquerci (changes requested):
Hello,

Thanks for your contribution.

On this PR you tackle too many things.

But I see that only this commit can be approved IMHO.

https://github.com/FriendsOfSymfony1/symfony1/pull/242/commits/74a7e89eb6c6a57c6a186769d45b20e1b937be68

terdia:
> approved

That's correct, I reverted the other one, through conflict resolution

terdia:
> Hello,
> 
> Thanks for your contribution.
> 
> On this PR you tackle too many things.
> 
> But I see that only this commit can be approved IMHO.
> 
> [74a7e89](https://github.com/FriendsOfSymfony1/symfony1/commit/74a7e89eb6c6a57c6a186769d45b20e1b937be68)

Thanks, I will create another or for it now.

PR #246, opened 2021-05-12
Update PHPDoc for sfWebController::redirect() func

cmdeviant:
We can actually pass array to sfWebController::redirect() function as first parameter, but IDE wrongly highlights this as "Expected parameter of type 'string', 'array' provided".

### DIFF
--- a/lib/controller/sfWebController.class.php
+++ b/lib/controller/sfWebController.class.php
@@ -168,7 +168,7 @@ abstract class sfWebController extends sfController
   /**
    * Redirects the request to another URL.
    *
-   * @param mixed $url         An associative array of URL parameters or an internal URI as a string
+   * @param string|array $url         An associative array of URL parameters or an internal URI as a string
    * @param int    $delay      A delay in seconds before redirecting. This is only needed on
    *                           browsers that do not support HTTP headers
    * @param int    $statusCode The status code
--- a/lib/controller/sfWebController.class.php
+++ b/lib/controller/sfWebController.class.php
@@ -168,7 +168,7 @@ abstract class sfWebController extends sfController
   /**
    * Redirects the request to another URL.
    *
-   * @param string $url        An associative array of URL parameters or an internal URI as a string
+   * @param mixed $url         An associative array of URL parameters or an internal URI as a string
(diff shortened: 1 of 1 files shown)


mentalstring (approved):

spirit-q2 (approved):

lib/controller/sfWebController.class.php
```diff
@@ -168,7 +168,7 @@ public function convertUrlStringToParameters($url)
   /**
    * Redirects the request to another URL.
    *
-   * @param string $url        An associative array of URL parameters or an internal URI as a string
+   * @param mixed $url         An associative array of URL parameters or an internal URI as a string
```
e1himself, line 171: > We can actually pass array [...] as first parameter

How about changing it to `string|array` instead then? 

The more precise the types are, the better IDE will help preventing errors.

cmdeviant: > How about changing it to string|array instead then?

Good point @e1himself. Originally I noticed that `redirect()` funciton has _string_ type for its **$url** parameter. But that differs from same _mixed_ parameter, named **$parameters**, in  [getUrl() function](https://github.com/FriendsOfSymfony1/symfony1/blob/10e1a1b175d4c86f44b9843d413b3cc2702ad5a8/lib/controller/sfWebController.class.php#L24-L32). So idea was to much their types, because what was passed into `redirect()` then will be passed into `genUrl()`. Looks like `genUrl()` PHPDoc should also be changed accordingly with what you suggest then. What do you think?

thirsch: In that case I wouldn't change too much of the old code. Thanks for the explanation.

thirsch (approved):

lib/controller/sfWebController.class.php
e1himself: > In that case I wouldn't change too much of the old code.

As for me, I'm both hands for improving phpDocs in the old code. 
I've fixed a lot of these in 2017: #153.

alquerci (approved):

e1himself (approved):
:+1:

thePanz (approved):

thePanz:
@cmdeviant Thans, merged!

Issue #125, opened 2016-05-14
sfAPCCache fail


zkiiito:
The tests on travis run without apc enabled for cli, so the sfAPCCacheTest just skips itself.
If I enable it in php.ini, they fail, as you can see [here](https://travis-ci.org/ingatlancom/symfony1/builds/130187838)

mentalstring:
Couldn't reproduce it with https://github.com/ingatlancom/symfony1/tree/sfAPCCache on any PHP version I have at hand. Even tried with PHP 5.3.29 like travis, but worked fine.

I suspect some APC configuration quirkiness so tried using the settings from travis [apc.ini](https://github.com/rezzza/travis-ci/blob/master/php/conf.d/apc.ini) but still no luck.

Is anyone able to reproduce this outside of travis?

zkiiito:
Is 
`apc.enable_cli = 1` 
in your php.ini? Without that, the tests are skipped.

mentalstring:
Yes,

```
sfAPCCache $ php -n -dextension=apc.so -d"apc.enable_cli=1" test/unit/cache/sfAPCCacheTest.php
1..64
# ->initialize()
# ->set() ->get() ->has()
ok 1 - ->set() returns true if data are stored in cache
ok 2 - ->get() retrieves data form the cache
ok 3 - ->has() returns true if the cache exists
ok 4 - ->set() takes a lifetime as its third argument
ok 5 - ->get() returns the default value if cache has expired
ok 6 - ->has() returns true if the cache exists
ok 7 - ->get() returns null if the cache does not exist
ok 8 - ->get() takes a default value as its second argument
ok 9 - ->has() returns false if the cache does not exist
ok 10 - ->set() overrides previous data stored in the cache
ok 11 - ->get() retrieves the latest data form the cache
ok 12 - ->set() false data are stored in the cache
ok 13 - ->has() returns true if the cache exists with false value
ok 14 - ->get() retrieves the latest data form the cache
ok 15 - ->get() does not return default value if false is stored
ok 16 - ->set() accepts a namespaced cache key
ok 17 - ->set() accepts a namespaced cache key
ok 18 - ->set() accepts a namespaced cache key
# ->clean()
ok 19 - ->clean() cleans old cache key if given the sfCache::OLD argument
ok 20 - ->clean() cleans old cache key if given the sfCache::OLD argument
ok 21 - ->clean() cleans all cache key if given the sfCache::ALL argument
[...]

Issue #48, opened 2014-03-09
sfAPCCache seems to not be compatible with apcu


j0k3r:
APCu seems to not use the same way to store "key".
sfAPCCache->getCacheInfo():

``` php
  protected function getCacheInfo($key)
  {
    if (!$this->enabled)
    {
      return false;
    }

    $infos = apc_cache_info('user');

    if (is_array($infos['cache_list']))
    {
      foreach ($infos['cache_list'] as $info)
      {
        if ($this->getOption('prefix').$key == $info['info'])
        {
          return $info;
        }
      }
    }

    return null;
  }
```

Before this if `if ($this->getOption('prefix').$key == $info['info'])`, I vardumped the `$info` array and here it is for APCu & APC.
With APCu:

```
array (size=9)
  'key' => string 'account:/localhost/all/main/index/__sf_cache_key/item_tags_8267' (length=63)
  'ttl' => int 86400
  'nhits' => float 0
  'mtime' => int 1394384708
  'ctime' => int 1394384708
  'dtime' => int 0
  'atime' => int 1394384708
  'ref_count' => int 0
  'mem_size' => int 1712
```

And this with APC:

```
array (size=9)
  'info' => string 'account:/localhost/all/main/index/__sf_cache_key/item_tags_8267' (length=63)
  'ttl' => int 86400
  'nhits' => float 0
  'mtime' => int 1394384708
  'ctime' => int 1394384708
  'dtime' => int 0
  'atime' => int 1394384708
  'ref_count' => int 0
  'mem_size' => int 1712
```

Spot the first key from the array ? info != key
Should we create a `sfAPCuCache` ? The problem will also appear in `removePattern()`.

mentalstring:
I have been using APCu in production without any issue with sfAPCCache

I believe APCu 4.0.7 [has fixed this](https://pecl.php.net/package-changelog.php?package=APCu&release=4.0.7) and with it, the API is the same as APC. Unless I'm missing something, this fix only makes it compatible with older versions of APCu.

Issue #261, opened 2021-12-18
PHP 8.1 support


mbrung:
Dear project maintainers, 

Thank you very much for all your hard work, it has helped me tremendously!

Has any work been carried out towards supporting PHP 8.1?  I haven't looked in details yet at what it would entail, just raising the question (and avoid duplicate work).

Thanks again!

axelusarov:
Same here. Would be great to know if there are any plans to fix PHP compatibility issues.

alquerci:
Hello @mbrung, @axelusarov 

As, I know, there are not any plans to fix PHP compatibility issues yet.

### A short view on CI

Regarding the [nightly build](https://travis-ci.org/github/FriendsOfSymfony1/symfony1/jobs/772313153) that run PHP 8.0.3-dev
There are a bunch of issues.
It seems that's not so easy.

### Canvas

* This project will need to keep the BC to PHP 5.3, so it will be harder and harder to comply with this constraint.

### Contributions

Personalty, I do not have any project running on sf1, and therefore I do not have time to work on it.
But, I am working on applications running on PHP 7, so it will be instructive to contribute to the support of this one.

I will be glade to help for code-review.

But, I do not know the availability of guys who are merging access. cc @thePanz, @j0k3r 

@mbrung, @axelusarov do you wish to contribute?

### A plan ?

But let's, it is not too late to start tackling one problem at a time.
The goal is to unscale the difficulty with tiny issue.

1. for each [Backward Incompatible Changes][0], needs to search code occurrences
2. etc...

### Idea

* Using issue or PR prefix `[PHP 8.0]` and `[PHP 8.1]`.
    Example: `[PHP 8.0] String to Number Comparison`

[0]: https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible

j0k3r:
I really think you need to invest time in upgrading your application to move away from symfony1 instead of upgrading your server.

e1himself:
@mbrung is your app under active development or is it just a legacy project requiring maintenance?

mentalstring:
From a quick look, it looks like it does need a bit of work to support PHP 8.x, while being particularly tricky to keep support for PHP 5.3 at the same time as the changes keep compounding. I think at some point, PHP itself will have BC breaks that will not allow this codebase to support all PHP versions.

[...]

PR #263, opened 2022-01-23
Fix APC tests and add APCu support

mentalstring:
The APC support has been broken for a while now (see [here](https://github.com/FriendsOfSymfony1/symfony1/issues/125) and [here](https://github.com/FriendsOfSymfony1/symfony1/issues/48)) but since `sfAPCCacheTest` is often skipped, it's been overlooked.

The issue is that APC has been dropped in more recent versions of PHP and the user cache part of it is now provided by APCu. However, the functions have been changed (eg: `apc_get()` to `apcu_get()`, and while there was a [compatibility layer](https://github.com/krakjoe/apcu-bc) for a while it doesn't seem to be maintained anymore. These days one needs to use APC or APCu, often as a requirement by the PHP being used. In other words, `sfAPCCache`, as is, no longer works in recent versions of PHP where only APCu is available.

This PR fixes the issues with APC inconsistencies when using with negative TTLs, and also adds a new sfAPCuCache that is a drop-in replacement for sfAPCCache. It's pretty minor changes from sfAPCCache, and while it is duplicated code, it's likely one of the cleanest ways to address this without overcomplicating things. `sfAPCCacheTest` loads the right one according to which extension is present (apc vs apcu).

Lastly, not sure if we should now drop the line bellow allow `travis.yml` to run the tests with APC (&APCu) again. 

https://github.com/FriendsOfSymfony1/symfony1/blob/6f521e9ef9e800c655313a5f21ac3f55552fbcf4/.travis.yml#L36

### DIFF
--- a/test/unit/cache/sfCacheDriverTests.class.php
+++ b/test/unit/cache/sfCacheDriverTests.class.php
@@ -18,7 +18,8 @@ class sfCacheDriverTests
     $t->is($cache->get('test'), $data, '->get() retrieves data form the cache');
     $t->is($cache->has('test'), true, '->has() returns true if the cache exists');
 
-    $t->ok($cache->set('test', $data, -10), '->set() takes a lifetime as its third argument');
+    $t->ok($cache->set('test', $data, 1), '->set() takes a lifetime as its third argument');
+    sleep(2);
     $t->is($cache->get('test', 'default'), 'default', '->get() returns the default value if cache has expired');
     $t->is($cache->has('test'), false, '->has() returns true if the cache exists');
 
@@ -47,21 +48,24 @@ class sfCacheDriverTests
     // ->clean()
     $t->diag('->clean()');
     $data = 'some random data to store in the cache system...';
-    $cache->set('foo', $data, -10);
+    $cache->set('foo', $data, 1);
+    sleep(2);
     $cache->set('bar', $data, 86400);
 
     $cache->clean(sfCache::OLD);
     $t->is($cache->has('foo'), false, '->clean() cleans old cache key if given the sfCache::OLD argument');
     $t->is($cache->has('bar'), true, '->clean() cleans old cache key if given the sfCache::OLD argument');
 
-    $cache->set('foo', $data, -10);
+    $cache->set('foo', $data, 1);
+    sleep(2);
     $cache->set('bar', $data, 86400);
 
     $cache->clean(sfCache::ALL);
     $t->is($cache->has('foo'), false, '->clean() cleans all cache key if given the sfCache::ALL argument');
     $t->is($cache->has('bar'), false, '->clean() cleans all cache key if given the sfCache::ALL argument');
 
-    $cache->set('foo', $data, -10);
+    $cache->set('foo', $data, 1);
+    sleep(2);
     $cache->set('bar', $data, 86400);
 
     $cache->clean();
@@ -128,7 +132,8 @@ class sfCacheDriverTests
       $t->ok($delta >= $lifetime - 1 && $delta <= $lifetime, '->getTimeout() returns the timeout time for a given cache key');
     }
 
-    $cache->set('bar', 'foo', -10);
+    $cache->set('bar', 'foo', 1);
+    sleep(2);
     $t->is($cache->getTimeout('bar'), 0, '->getTimeout() returns the timeout time for a given cache key');
 
     foreach (array(86400, 10) as $lifetime)
@@ -152,7 +157,8 @@ class sfCacheDriverTests
       $t->ok($lastModified >= time() - 1 && $lastModified <= time(), '->getLastModified() returns the last modified time for a given cache key');
     }
 
-    $cache->set('bar', 'foo', -10);
+    $cache->set('bar', 'foo', 1);
+    sleep(2);
     $t->is($cache->getLastModified('bar'), 0, '->getLastModified() returns the last modified time for a given cache key');
 
     foreach (array(86400, 10) as $lifetime)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,37 @@
 os: linux
-dist: xenial
+dist: focal
 language: php
 
 php:
-    - 7.0
-    - 7.1
-    - 7.2
-    - 7.3
-    - 7.4
-    - nightly
+  - 7.4
+  - 8.0
+  - 8.1
+  - nightly
 
 # run build against nightly but allow them to fail
 jobs:
     fast_finish: true
     allow_failures:
         - php: nightly
+        - php: 5.3
+        - php: 5.4
     include:
         - php: 5.3
           dist: precise
         - php: 5.4
-          dist: precise
+          dist: trusty
         - php: 5.5
           dist: trusty
         - php: 5.6
           dist: trusty
+        - php: 7.0
+          dist: xenial
+        - php: 7.1
+          dist: xenial
+        - php: 7.2
+          dist: xenial
+        - php: 7.3
+          dist: xenial
 
 services:
     - memcached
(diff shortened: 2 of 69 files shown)


test/unit/cache/sfCacheDriverTests.class.php
```diff
@@ -18,7 +18,8 @@ static public function launch($t, $cache)
     $t->is($cache->get('test'), $data, '->get() retrieves data form the cache');
     $t->is($cache->has('test'), true, '->has() returns true if the cache exists');
 
-    $t->ok($cache->set('test', $data, -10), '->set() takes a lifetime as its third argument');
```
alquerci, line 21: Not sure if we need to keep the support for negative lifetime.

Also, 0 means for APC forever.

What do your think?

mentalstring: At least from a sfAPCCache viewpoint, I don't see a particular use for negative TTLs (which from the APC side, is not a documented behavior as a way to invalidate a key).

Looking through the [docs](https://symfony.com/legacy), I didn't find any reference about using negative TTL  that could lead developers to trust that behavior.

Hence, I think negative TTLs were used here only for convenience of testing. This particular test is to make sure keys expire — not that negative TTLs have that effect. One could actually argue that the use of sleep() is actually a better way to test that keys do expire over time using a regular (eg: positive) TTL. Negative TTLs sure are convenient for testing, and allow faster test runs, but positive TTLs and sleep() actually makes sure keys expire with the most common use (positive TTLs).

In any case, mind we are not changing sfAPCCache behavior here, so there's no BC break here. The only change is the way we test key expiration.

alquerci (commented):
LGTM

This patch is no-op until developers change the configuration.

* What about the factory configuration to update by developers?
* What about doing an iteration on the enabled switch as it was added to fix the issue (perhaps) that this PR is fixing (e.g. 5a0adceb79d44b76f3ad7d39d041133fafe065f7) ? cc @jeromemacias 
* Who is responsible to choose the good APC implementation, the framework or developers?

alquerci (changes requested):
Tests failed at least on PHP 7.0

```
                                                                       
  Error: Class name must be a valid object or a string                 
  (in test/unit/cache/sfAPCCacheTest.php on line 26)                   
                                                                       
```

mentalstring:
>   Error: Class name must be a valid object or a string                 
>   (in test/unit/cache/sfAPCCacheTest.php on line 26)                   

Should be good now. @alquerci You might want to set all APCU_VERSION on docker-compose.yml now to get this to run.

> * What about the factory configuration to update by developers?
> * Who is responsible to choose the good APC implementation, the framework or developers?

The choice between apc or apcu always needs to be a deliberate choice by the developer (according to what's available on the environment), just like they could choose to use `sfMemcacheCache` instead.

> * What about doing an iteration on the enabled switch as it was added to fix the issue (perhaps) that this PR is fixing (e.g. [5a0adce](https://github.com/FriendsOfSymfony1/symfony1/commit/5a0adceb79d44b76f3ad7d39d041133fafe065f7)) ? cc @jeromemacias is it?

I think it would mostly just add complexity with little or no benefit? Personally I find the current approach to be the cleanest, as one decides to use one or the other (eg: on factories) without having to introduce a new setting.

Then again, it's not clear to me why that enable switch was added to sfAPCCache as there's already a sfNoCache driver for that purpose.

test/unit/cache/sfCacheDriverTests.class.php
```diff
@@ -18,7 +18,8 @@ static public function launch($t, $cache)
...
+    $t->ok($cache->set('test', $data, 1), '->set() takes a lifetime as its third argument');
+    sleep(2);
```
alquerci, line 22: About the test duration concern.

number of sleep * sleep duration * number of class
6 * 2 * 4 = 48 seconds is added to the test suite duration, this not count all the matrix with all PHP versions.

I suggest trying to find a way to mitigate this issue.

Does it be possible to use at most 2 sleep per class test (e.g. 2 * 2 * 4 = 16) ?

The test will need to be refactored, for sure.

I may help if you need.

alquerci (changes requested):

alquerci (changes requested):
Test failed, can you reproduce?

test/bin/test php72 highest

APC_ENABLE_CLI=on

```
                                                                       
  unit/storage/sfCacheSessionStorageTest.php                           
                                                                       

#   at unit/storage/sfCacheSessionStorageTest.php line 56
>   ->regenerate() regenerated the session with a different session id
      ''
          ne
      ''
#   at unit/storage/sfCacheSessionStorageTest.php line 60
>   ->regenerate() regenerated the session with a different session id
      ''
          ne
      ''
```

test/unit/cache/sfCacheDriverTests.class.php
mentalstring: Not sure whether this is worth our time optimising for CI time on a framework that is not use used out there?

I mean, I'm all for optimising code performance for production, but I don't think Travis will be bothered if it runs another minute or 5 and the scenario where a dev is manually testing **all** tests with **all** PHP versions is somewhat limited: in most cases, one is manually running a single unit test that failed on a specific version (or skipping most using the `--only-failed` switch).

Personally I would go for not touching the testing logic here even further and save that time to invest on the [main reason](https://github.com/FriendsOfSymfony1/symfony1/issues/261) we are touching this.

alquerci: > I don't think Travis will be bothered if it runs another minute or 5

Just a reminder, Travis-CI is the past as it became paying.

So it is the responsibility, to dev to run all tests, and we know them does not have much time to work on this.

Okay, let's say it's not required, for now, to invest effort to avoid slowing down the test duration.

alquerci:
Nice investigation.

Since PHP 7.2 this warning is triggered

```
Warning: session_id(): Cannot change session id when headers already sent in /app/lib/storage/sfCacheSessionStorage.class.php on line 262
```

> I would still prefer sfFileCache was used here

It seems a good track, but apply this modification does not fix the issue.

mentalstring:
> It seems a good track, but apply this modification does not fix the issue.

Yes, replacing the use of `sfAPCCache` doesn't fix the test — it's just an additional improvement to reduce dependencies on something (APC) that is not being tested. 

The proposed solution is to use output buffering with `ob_start()` — I tried and it works. What do you think of the idea?

alquerci:
> The proposed solution is to use output buffering with ob_start() — I tried and it works. What do you think of the idea though?

Yeah, let's go it.

jeromemacias:
>     * What about doing an iteration on the enabled switch as it was added to fix the issue (perhaps) that this PR is fixing (e.g. [5a0adce](https://github.com/FriendsOfSymfony1/symfony1/commit/5a0adceb79d44b76f3ad7d39d041133fafe065f7)) ? cc @jeromemacias is it?

Sorry, I don't remember exactly why I did this change, but I think it's hard to be sure at 100% that the environement have APC installed and well configured so in some cases we cannot tell to use sfNoChache adapter. I's easier for me to say "Use APC if available".

By the way, this PR looks good to me.

alquerci:
> Sorry, I don't remember exactly why I did this change, but I think it's hard to be sure at 100% that the environement have APC installed and well configured so in some cases we cannot tell to use sfNoChache adapter. I's easier for me to say "Use APC if available".

Alright.

For comparison with latest Symfony cache version.

https://github.com/symfony/symfony/blob/4274a1f/src/Symfony/Component/Cache/Adapter/ApcuAdapter.php#L30

alquerci (changes requested):
On this PR there are 2 commits that turn it to un-mergeable state.

* Add support for SameSite cookie session setting => #215
* PHP8.0 & 8.1 support => #266

mentalstring:
Closing this in favor of #267 to get it from a different branch.

PR #264, opened 2022-02-24
Accept PHP 8.1's new $_FILES['full_path'] member

gileri:
[PHP 8.1 introduced a new member to the $_FILES superglobal](https://php.watch/versions/8.1/$_FILES-full-path).

This PR allows to accept both array members.


Tybaze (changes requested):
There is a Notice with PHP < 8.1 : 
`Notice: Undefined index: full_path in ... sfWebRequest.class.php:897
`
The new index full_path is only available after 8.1 😃

alquerci:
@Tybaze nice catch.

Did you see it from the test suite?

I wonder if it is covered.

Tybaze:
@alquerci , No, just reviewing the PR, got a doubt, make a custom test on an 7.4 env and trigger the error.

We should create a specific "adaptive" test for 8.1+ and before, 
And as it is a "Notice" I don't think it will crash the test

PR #262, opened 2022-01-22
Add consistent local environment for testing with docker and docker-compose

alquerci:
Goal
----

* Removes all side work for contributors that do not provide any value, except effective improvements
* Provide the same testing environment for everyone.
* Help contributors to win the big challenge of PHP new versions supports, while BC layer are kept.
* Provide a bit more confidence during the development by reducing the testing feedback loop.

Usage
-----

Look at the README.md.

In short
```
test/bin/test
```

Test results
----------

### Same matrix than travis-ci

| PHP version | Dependency | Status
| ---- | ---- | ----
| php53 | highest | :green_circle:
| php54 | highest | :green_circle:
| php55 | highest | :green_circle:
| php56 | highest | :green_circle:
| php70 | highest | :green_circle:
| php71 | highest | :green_circle:
| php72 | highest | :green_circle:
| php73 | highest | :red_circle: 
| php74 | highest | :red_circle: 

```
#   at [sfDoctrine] functional/AdminGenTest.php line 17
>   response selector "select" matches "NewActivePendingExpired"
           got: '
New
Active
Pending
Expired
'
      expected: 'NewActivePendingExpired'
```

### New additional matrix (disabled by default)

| PHP version | Dependency | Status
| ---- | ---- | ----
| php53 | lowest | :green_circle:
| php54 | lowest | :green_circle:
| php55 | lowest | :green_circle:
| php56 | lowest | :green_circle:
| php70 | lowest | :green_circle:
| php71 | lowest | :green_circle:
| php72 | lowest | :green_circle:
| php73 | lowest | :red_circle: 
| php74 | lowest | :red_circle: 
| php80 | highest | :red_circle: 
| php81 | highest | :red_circle: 
| php80 | lowest | :red_circle: 
| php81 | lowest | :red_circle: 

Related to #261

PR #265, opened 2022-03-18
PHP8

teymour:
PHP8 corrections

### DIFF
--- a/lib/addon/sfPager.class.php
+++ b/lib/addon/sfPager.class.php
@@ -529,7 +529,7 @@ abstract class sfPager implements Iterator, Countable
    *
    * @see Iterator
    */
-  public function current()
+  public function current(): mixed
   {
     if (!$this->isIteratorInitialized())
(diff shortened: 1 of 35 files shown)


alquerci:
Hello @teymour 

That's a huge work here.

Can you base your work on #262 and #263 ?

Tybaze (commented):
Please be carefull,

date and strftime do not share the same format.

You can change the default timeFormat, but it will break a lot of setup (because default Sf1 timeformat is missing year, it's common to change it)

I'll try to provide a fix

Tybaze:
I think we can close this PR, everything has been included into the merged #266 

Thank you very much @teymour your list of modifications was usefull 👍

teymour:
> Thank you very much @teymour your list of modifications was usefull +1

Thanks for your work ! This new version works on our projects on PHP 8.1 and 7.4. It's a great news !!

PR #267, opened 2022-08-03
Fix APC tests and add APCu support

mentalstring:
This replaces #263.

The APC support has been broken for a while now (see [here](https://github.com/FriendsOfSymfony1/symfony1/issues/125) and [here](https://github.com/FriendsOfSymfony1/symfony1/issues/48)) but since `sfAPCCacheTest` is often skipped, it's been overlooked.

The issue is that APC has been dropped in more recent versions of PHP and the user cache part of it is now provided by APCu. While there was a [compatibility layer](https://github.com/krakjoe/apcu-bc) it doesn't seem to be maintained anymore, so the functions now have different names (eg: `apc_get()` to `apcu_get()`. These days, according to the PHP version use, one needs to use either APC or APCu. In other words, `sfAPCCache`, as is, no longer works in recent versions of PHP where only APCu is available.

This PR fixes some APC/APCu inconsistencies when using negative TTLs in the tests, and also adds a new sfAPCuCache that is a drop-in replacement for sfAPCCache. It's pretty minor changes from sfAPCCache, and while it is duplicated code, I think it's one of cleanest ways to address this without overcomplicating things. `sfAPCCacheTest` loads the right one according to which extension is present (apc vs apcu).

Lastly, not sure if we should now drop the line bellow allow `travis.yml` to run the tests with APC (&APCu) again. 

https://github.com/FriendsOfSymfony1/symfony1/blob/6f521e9ef9e800c655313a5f21ac3f55552fbcf4/.travis.yml#L36

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -3,9 +3,3 @@
 #
 # Copy to `.env` in order to use it.
 #
-
-# APC test are disabled.
-#
-# To enable them in order to provide a fix, set to "on".
-#
-APC_ENABLE_CLI=off
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -39,6 +39,8 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php-version }}"
+          extensions: apcu
+          ini-values: apc.enable_cli=1
 
       - name: Get composer cache directory
         id: composer-cache
--- a/data/bin/check_configuration.php
+++ b/data/bin/check_configuration.php
@@ -77,7 +77,7 @@ check(function_exists('utf8_decode'), 'The utf8_decode() is available', 'Install
 check(function_exists('posix_isatty'), 'The posix_isatty() is available', 'Install and enable the php_posix extension (used to colorized the CLI output)', false);
 
 $accelerator =
-  (function_exists('apc_store') && ini_get('apc.enabled'))
+  ((function_exists('apc_store') || function_exists('apcu_store')) && ini_get('apc.enabled'))
   || function_exists('eaccelerator_put') && ini_get('eaccelerator.enable')
   || function_exists('xcache_set');
 check($accelerator, 'A PHP accelerator is installed', 'Install a PHP accelerator like APC (highly recommended)', false);
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -94,7 +94,7 @@ services:
       args:
         PHP_TAG: '7.0-cli-jessie'
         MEMCACHE_VERSION: '4.0.5.2'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
   php71:
     <<: *services_php54
@@ -103,7 +103,7 @@ services:
       args:
         PHP_TAG: '7.1-cli-jessie'
         MEMCACHE_VERSION: '4.0.5.2'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   php72:
@@ -113,7 +113,7 @@ services:
       args:
         PHP_VERSION: '7.2'
         MEMCACHE_VERSION: '4.0.5.2'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   php73:
@@ -123,7 +123,7 @@ services:
       args:
         PHP_VERSION: '7.3'
         MEMCACHE_VERSION: '4.0.5.2'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   php74:
@@ -133,7 +133,7 @@ services:
       args:
         PHP_VERSION: '7.4'
         MEMCACHE_VERSION: '4.0.5.2'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   php80:
@@ -143,7 +143,7 @@ services:
       args:
         PHP_VERSION: '8.0'
         MEMCACHE_VERSION: '8.0'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   php81:
@@ -153,7 +153,7 @@ services:
       args:
         PHP_VERSION: '8.1'
         MEMCACHE_VERSION: '8.0'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
   php82:
     <<: *services_php54
@@ -162,7 +162,7 @@ services:
       args:
         PHP_VERSION: '8.2'
         MEMCACHE_VERSION: '8.0'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
   php83:
     <<: *services_php54
@@ -171,7 +171,7 @@ services:
       args:
         PHP_VERSION: '8.3'
         MEMCACHE_VERSION: '8.0'
-        APCU_VERSION: ''
+        APCU_VERSION: '5.1.23'
 
 
   db:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
           echo 'short_open_tag = off'
           echo 'magic_quotes_gpc = off'
           echo 'date.timezone = "UTC"'
-          echo 'apc.enable_cli = ${APC_ENABLE_CLI-off}'
+          echo 'apc.enable_cli = on'
           echo 'apc.use_request_time = 0'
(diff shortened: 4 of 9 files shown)


thePanz (approved):
LGTM

Maybe @alquerci (as you reviewed the original PR) would give a feedback too?

alquerci (approved):
LGTM

:heavy_check_mark: Tests passes with `apcu-5.1.21`
:red_circle: But not with `apcu-5.1.22`,

---

I come back since long time, so I went to run all tests.

I got on PHP 7.0 and lower.
```
Parse error: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE) in /app/lib/log/sfLogger.class.php on line 30
```

It is not a blame, I just expose what I see.

For PHP >7.0 all tests passes, good job :heavy_check_mark: .

mentalstring:
> ✔️ Tests passes with `apcu-5.1.21`
> 🔴 But not with `apcu-5.1.22`,

I believe the issue is a not quite planned [behaviour change](https://github.com/krakjoe/apcu/issues/451) on 5.1.22 that even broke the apc.php info page. This has been [reverted](https://github.com/krakjoe/apcu/commit/8c05c515b698d5f95555a6afdc00846405d50b40) on 5.1.23 released a month ago — the test passes again with that.

I don't think it's worth putting a complicated workaround for a specific apcu version — maybe we can just skip() if it's `apcu-5.1.22`?


> I got on PHP 7.0 and lower.
> 
> ```
> Parse error: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE) in /app/lib/log/sfLogger.class.php on line 30
> ```

I'm having trouble running the tests from master.
```
$ docker --version
Docker version 24.0.6, build ed223bc
$ docker-compose --version
Docker Compose version v2.23.0-desktop.1

$ test/bin/test php73 highest test/unit/cache/sfAPCCacheTest.php
+ docker-compose build
failed to solve: process "/bin/sh -c set -eux;     apt-get update;   apt-get install -y     curl     autoconf2.13   ;   rm -r /var/lib/apt/lists/*;     curl -sSLfO http://launchpadlibrarian.net/140087283/libbison-dev_2.7.1.dfsg-1_amd64.deb;   curl -sSLfO http://launchpadlibrarian.net/140087282/bison_2.7.1.dfsg-1_amd64.deb;   dpkg -i libbison-dev_2.7.1.dfsg-1_amd64.deb;   dpkg -i bison_2.7.1.dfsg-1_amd64.deb;   rm *.deb;     curl -sSLf \"https://php.net/get/php-$PHP_VERSION.tar.bz2/from/this/mirror\" -o php.tar.bz2;   echo 'c4e1cf6972b2a9c7f2777a18497d83bf713cdbecabb65d3ff62ba441aebb0091  php.tar.bz2' | sha256sum -cw --status;     mkdir -p /usr/src/php;   tar -xf php.tar.bz2 -C /usr/src/php --strip-components=1;   rm php.tar.bz2*;     cd /usr/src/php;   ./buildconf --force;   ./configure --disable-cgi     $(command -v apxs2 > /dev/null 2>&1 && echo '--with-apxs2' || true)     --with-pdo-mysql     --with-zlib     --enable-mbstring   ;   make -j\"$(nproc)\";   make install;     dpkg -r     bison     libbison-dev   ;   apt-get purge -y --auto-remove     autoconf2.13   ;   rm -r /usr/src/php" did not complete successfully: exit code: 100
```


Any idea, @alquerci?

alquerci:
Hello @mentalstring 

I agree, for the APCu version.

For running tests.

Try with patches.
- https://github.com/FriendsOfSymfony1/symfony1/pull/270
- https://github.com/FriendsOfSymfony1/symfony1/pull/293

mentalstring:
I've rebased and added the skip for APCu 5.1.22. Also filled `APCU_VERSION` to 5.1.23 on `docker-compose.yml` so that the APCu tests can run.

I think it's ready for merging. But if the minimum version becomes PHP 7.0, then I suppose instead of having two classes (`sfAPCCache.class.php` and `sfAPCuCache.class.php`), we could just change the original `sfAPCCache.class.php` to use the new APCu API. /cc @thePanz 


> I got on PHP 7.0 and lower.
> 
> ```
> Parse error: syntax error, unexpected 'const' (T_CONST), expecting variable (T_VARIABLE) in /app/lib/log/sfLogger.class.php on line 30
> ```
> 
> It is not a blame, I just expose what I see.
> 
> For PHP >7.0 all tests passes, good job ✔️ .

Still can't reproduce this. I have tested `test/unit/cache/sfAPCCacheTest.php` with all PHP versions and also ran the whole test suite with PHP 7.0 and didn't have any problem. Can you still reproduce it @alquerci ?

thirsch (approved):

alquerci:
@mentalstring What steps did you done to execute the whole test suite with PHP 7.0?

I suppose
1. You apply patches
- https://github.com/FriendsOfSymfony1/symfony1/pull/270
- https://github.com/FriendsOfSymfony1/symfony1/pull/293
2. Execute `test/bin/test --php-versions php70`

.env.dist
```diff
@@ -8,4 +8,4 @@
 #
 # To enable them in order to provide a fix, set to "on".
 #
-APC_ENABLE_CLI=off
+APC_ENABLE_CLI=on
```
alquerci, line 11: Still exists a reason to disable APC test as this PR provide the fix.

So I suggest removing this flag, and always enable it.

mentalstring: No sure I follow. The point of this PR is precisely to fix the issues that have been preventing the APC tests from running as they should. What else is missing?

I would actually suggest removing that setting altogether and resume always running the APC tests as I don't see the need for the setting once the APC tests are fixed (by this PR). If APC/APCu is not installed, the tests just skip.

mentalstring:
> @mentalstring What steps did you done to execute the whole test suite with PHP 7.0?
> 
> I suppose
> 
> 1. You apply patches
> 
> * [Fix docker images build to handle old linux distributions #270](https://github.com/FriendsOfSymfony1/symfony1/pull/270)
> * [Restore support of PHP <7.1 #293](https://github.com/FriendsOfSymfony1/symfony1/pull/293)
> 
> 2. Execute `test/bin/test --php-versions php70`

Yep, pretty much. The test completed with no problem for me.

.env.dist
alquerci: > I would actually suggest removing that setting altogether and resume always running the APC tests

Yes, you can do it.

alquerci:
After removing `APC_ENABLE_CLI` setting, for me, this PR can be merged.

---
@mentalstring without this following patch, you will get the `unexpected 'const'` parse error.

- https://github.com/FriendsOfSymfony1/symfony1/pull/293

But this issue is unrelated with this PR, except for the minimum PHP version supported.
As you said, no need to support APC is the minimum PHP version supported is at least 7.0.

.env.dist
mentalstring: I've removed `APC_ENABLE_CLI` and have set `apc.enable_cli=1` so that the APC tests always run.

mentalstring:
> After removing `APC_ENABLE_CLI` setting, for me, this PR can be merged.

Should be ready now.

alquerci (approved):

thePanz:
Thanks @mentalstring !
Would you please:
- use the conventional commit messages (just a nitpick, not so strongly enforcing it tho)
- check if the github workflow pipeline should be updated to *always* enable the apcu extension in the "php setup" action: this would make sure that we run the tests with the extension

wdyt?

mentalstring:
> * check if the github workflow pipeline should be updated to _always_ enable the apcu extension in the "php setup" action: this would make sure that we run the tests with the extension

Ah, good point. I think I managed to enable APCu on the CI runs, and I checked that it already installs 5.1.23 so that the test doesn't get skipped.


> * use the conventional commit messages (just a nitpick, not so strongly enforcing it tho)

Happy to, just not sure if I covered everything... Was it just the line length?

thePanz:
> Ah, good point. I think I managed to enable APCu on the CI runs, and I checked that it already installs 5.1.23 so that the test doesn't get skipped.
 Nice! Thanks!

> Happy to, just not sure if I covered everything... Was it just the line length?
Fine as it is now! I was more for a prefix with "Fix" :) all good now!

One of the commit's long-message reads
>  as an alternative
to APC which longer works with recent versions of PHP.

It should read "..which *no* longer works..." instead (the *no* is missing)

thePanz:
Merged, thanks @mentalstring ! :tada:

PR #277, opened 2023-01-12
Fix(deprecations) Fix deprecations in sfMessageSource_Aggregate,

thePanz:
add return types from implemented sfIMessageSource


thirsch:
What exactly is the deprecations fix here?

thePanz:
@thirsch context: I am wrapping a legacy SF 1.5 application, within a new SF application (using a dedicated Controller to wrap all legacy paths).

I am getting deprecation logs as the following:
```
[2023-01-12T14:06:28.422078+00:00] deprecation.INFO: User Deprecated: Method "sfIMessageSource::append()" might add "void" as a native return type declaration in the future. Do the same in implementation "sfMessageSource_Aggregate" now to avoid errors or add an explicit @return annotation to suppress this message. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Method \"sfIMessageSource::append()\" might add \"void\" as a native return type declaration in the future. Do the same in implementation \"sfMessageSource_Aggregate\" now to avoid errors or add an explicit @return annotation to suppress this message. at /srv/app/vendor/symfony/error-handler/DebugClassLoader.php:330)"} []
```

The reason: the `sfMessageSource_Aggregate` class does not use the same phpdoc as `sfIMessageSource`.
Adding them, removes the deprecation logs

thirsch:
Thanks, I wasn't aware, that Symfony is also parsing the PhpDoc-Blcok and expected at least some PHP 8 Attributes. ;-)

thirsch (approved):

Issue #276, opened 2023-01-10
php-cs-fixer workflow


connorhu:
As we talked about it, we could make a github action to test php coding standards (later maybe code sniffer).

This is a tested config for push event:
https://github.com/FriendsOfSymfony1/doctrine1/compare/master...connorhu:doctrine1:feature/phpcsfixer
https://github.com/FriendsOfSymfony1/symfony1/compare/master...connorhu:symfony1:feature/phpcsfixer

I have to test it for PR event, but it works for push. When I have tested for PR I will make a pull request. The cs-fixer setting is left at minimum (only tests what @thePanz noticed to in my PR at https://github.com/FriendsOfSymfony1/doctrine1/pull/94 ).
It's another question of what the setting should be at all. @Symfony ruleset would be logical, but it's a very aggressive and generates a very different style from current code formatting.

We should only test the files we touch.

thePanz:
Fixed in https://github.com/FriendsOfSymfony1/symfony1/pull/283

PR #280, opened 2023-02-14
Add(cs-fixer) Add php-cs-fixer configs and github action

thePanz:
Ref: https://github.com/FriendsOfSymfony1/symfony1/issues/276

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,44 @@
+on: [push, pull_request]
+name: Main
+jobs:
+  php-cs-fixer:
+    name: PHP-CS-Fixer
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Get Changed Files @ all
+        id: all-changes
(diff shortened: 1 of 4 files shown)


thirsch (approved):
lgtm!

Just two thoughts about it: How about ending a file with line break and as the .php-cs-fixer.dist.php is included within it's own defined finder, should the detector in step 2 (Get Changed Files @ all) detect changes there as well?

thePanz:
@thirsch 
> Just two thoughts about it: How about ending a file with line break and as the .php-cs-fixer.dist.php is included within it's own defined finder, should the detector in step 2 (Get Changed Files @ all) detect changes there as well?

I took the code from https://github.com/FriendsOfSymfony1/symfony1/issues/276 and just created a PR :)
To answer your question @thirsch :
1. yes, I would merge the two files in one, `.php-cs-fixer.dist.php`
2. gitworkflow: I would keep it simple, and not early-optimize (keep the simple configuration as shown in: https://github.com/OskarStark/php-cs-fixer-ga#usage-with-cache )

For further iterations:
- see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/usage.rst#using-php-cs-fixer-on-ci

thePanz:
@thirsch  @connorhu I would close this PR and instead work on https://github.com/FriendsOfSymfony1/symfony1/pull/283 WDYT?

PR #281, opened 2023-02-14
Fix PHP8.1 sfForm - Unsupported operand types: array + null on sfForm

Tybaze:
Fix PHP8.1 sfForm - Unsupported operand types: array + null on sfForm::updateValues()

Avoid error :
Fatal error: Uncaught TypeError: Unsupported operand types: array + null in lib\form\sfForm.class.php:319

Step to reproduce :
$form = new sfForm();
$form->updateValues(array('foo' => 'value'));

PHP 5.3 coding style

May use sfForm::getValues(), but as sfForm::updateValues() is a hack, it could be used before bind ($this->isBound = true);

### DIFF
--- a/lib/form/sfForm.class.php
+++ b/lib/form/sfForm.class.php
@@ -47,7 +47,7 @@ class sfForm implements ArrayAccess, Iterator, Countable
     protected $isBound = false;
     protected $taintedValues = array();
     protected $taintedFiles = array();
-    protected $values;
+    protected $values = array();
     protected $defaults = array();
     protected $fieldNames = array();
     protected $options = array();


lib/form/sfForm.class.php
```diff
@@ -316,7 +316,7 @@ public function isBound()
    */
   public function updateValues(array $values)
   {
-    $this->values = $values + $this->values;
+    $this->values = $values + (isset($this->values) ? $this->values : array());
```
thirsch, line 319: How about setting the initial value of $values to array()? https://github.com/FriendsOfSymfony1/symfony1/blob/c8cadb413101949316b0759510a1125580259647/lib/form/sfForm.class.php#L48

thePanz:
@Tybaze I updated this PR according to the comments (and rebased on the latest `master` branch)
Let me know if this fixes the error on your side and can be merged

thePanz (approved):

connorhu (approved):

PR #282, opened 2023-02-15
Apply explicit string casts for strtolower() and urlencode() functions

sebsnake:
(cherry picked from commit ebb7c6c3f780feb9c3e69caf3dbafcc4cea75f4d)

While we were migrating our application to PHP 8.1, some of our phpunit tests failed, as they somehow managed to get "null" inside symfony internals, where now strings are the only allowed argument types (resulting in deprecation warnings for now, but we configured our test suites to fail on those as well). 

This happend for cases of strtolower() and urlencode() functions. Thus, we simply got our application (test suites) working again by applying explicit string casts to those function calls via simple search-and-replace statements throught all source files. I am aware that this results in possibly many non-required casts, but we wanted to be sure and thus applied the patch to all uses of these functions.

Thought this might be useful for you as well, thus I made this pull request.

### DIFF
--- a/lib/action/sfAction.class.php
+++ b/lib/action/sfAction.class.php
@@ -404,7 +404,7 @@ abstract class sfAction extends sfComponent
    */
   public function getSecurityValue($name, $default = null)
   {
-    $actionName = strtolower($this->getActionName());
+    $actionName = strtolower((string) $this->getActionName());
 
     if (isset($this->security[$actionName][$name]))
     {
@@ -517,7 +517,7 @@ abstract class sfAction extends sfComponent
    */
   public function setViewClass($class)
   {
-    sfConfig::set('mod_'.strtolower($this->getModuleName()).'_view_class', $class);
+    sfConfig::set('mod_'.strtolower((string) $this->getModuleName()).'_view_class', $class);
   }
 
   /**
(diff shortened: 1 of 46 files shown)


mentalstring:
Seems a bit of a heavy solution which negates most of the advantages of having type checks: sometimes the nulls shouldn't be getting in there in the first place and the issue may be on the inputs: with the overall casting, we would miss the warnings/errors when on that.

I've been using the latest release on a medium usage website and only spotted one deprecation warning at:

https://github.com/FriendsOfSymfony1/symfony1/blob/2c5a4d625d1853cb207dfbb73817505831874371/lib/validator/sfValidatorError.class.php#L99

where `$value` sometimes is null when using `sfValidatorSchemaCompare` but I haven't manage to isolate the issue into a trivial trigger, so haven't go around submitting a fix.

Anyway, that's the only place that I've found a deprecation warning so far. Personally I wouldn't apply a blanket casting.

thirsch:
I wouldn't add it to every method as well. As discussed in https://github.com/FriendsOfSymfony1/symfony1/pull/283, we are about to move to a higher platform minimum version (7.4 as of latest discussions) and therefore a future change might bring typehints to the method signatures, bringing type errors back to who is responsible for calling them correctly.

connorhu:
I am using the latest version of symfony1 under 8.2 without any deprecation warnings. It is very strange that so many errors were found. If a method does not cause an error, it should not be changed.

PR #283, opened 2023-02-15
Add(linting) Add code linting to GitHub workflow, use php-cs-fixer

thePanz:
Fixing the php-cs-fixer is not done, as it would generate tons of changes

@thirsch @connorhu : WDYT of those sets?

Unfortunately the array syntax must be "long" as we support php v5.3 → would it make sense to only support v5.4 onward?


connorhu:
https://github.com/FriendsOfSymfony1/symfony1/issues/261#issuecomment-1019175462

Based on this table we can drop 5.3 and mark 5.4 as the minimum at next version.

My opinion is that everything should be dropped below 5.6.

thirsch:
From my point of view, every unsupported version could be dropped, as already mentioned in the linked issue and due to the fact, that we are only adding support for newer php versions. One could simply use an older version of sf1, if php 5.x is a must.

thePanz:
I would then propose to:

0. apply the code changes (keeping the long array syntax)
1. tag a latest release with 1.5.x → this will be the last release supporting php v5.x
2. create a branch `1.5.x` out of `master`
3. rename the `master` branch to `1.6.x`
4. Remove php v5.x from the supported versions
5. apply the php-cs-fixer introducing the short-array syntax (this way both branches are "aligned" on the coding styles)

I suggest to keep php v7.4 as min version to still support OS not yet shipping with php 8.x: this would allow a painless upgrade for those servers, while gaining some new language features

WDYT @thirsch @connorhu ?

thirsch:
> I would then propose to:
> 
> 0. apply the code changes (keeping the long array syntax)
> 1. tag a latest release with 1.5.x → this will be the last release supporting php v5.x
> 2. create a branch `1.5.x` out of `master`
> 3. rename the `master` branch to `1.6.x`
> 4. Remove php v5.x from the supported versions
> 5. apply the php-cs-fixer introducing the short-array syntax (this way both branches are "aligned" on the coding styles)
> 
> I suggest to keep php v7.4 as min version to still support OS not yet shipping with php 8.x: this would allow a painless upgrade for those servers, while gaining some new language features
> 
> WDYT @thirsch @connorhu ?

I'd keep master (or rename it to main) instead of a version name as the current working branch (step 3), but I agree with all other steps.

connorhu:
> > I would then propose to:
> > 
> > 0. apply the code changes (keeping the long array syntax)
> > 1. tag a latest release with 1.5.x → this will be the last release supporting php v5.x
> > 2. create a branch `1.5.x` out of `master`
> > 3. rename the `master` branch to `1.6.x`
> > 4. Remove php v5.x from the supported versions
> > 5. apply the php-cs-fixer introducing the short-array syntax (this way both branches are "aligned" on the coding styles)
> > 
> > I suggest to keep php v7.4 as min version to still support OS not yet shipping with php 8.x: this would allow a painless upgrade for those servers, while gaining some new language features
> > WDYT @thirsch @connorhu ?
> 
> I'd keep master (or rename it to main) instead of a version name as the current working branch (step 3), but I agree with all other steps.

I'd keep the master (or main) branch too, but the other steps looks good to me. If you specify the minimum version for the 1.6.x branch I can go with that in my Next Big Thing branch (spoiler: https://github.com/FriendsOfSymfony1/symfony1/compare/master...connorhu:symfony1:feature/phpunit )

thePanz:
> I'd keep the master (or main) branch too, but the other steps looks good to me. If you specify the minimum version for the 1.6.x branch I can go with that in my Next Big Thing branch (spoiler: [master...connorhu:symfony1:feature/phpunit](https://github.com/FriendsOfSymfony1/symfony1/compare/master...connorhu:symfony1:feature/phpunit) )

Wow, that's impressive! :+1: 
I guess my try to introduce phpstan and rector should wait a bit :)

connorhu:
@thePanz If you change the indent size in files (which is very fine by me) you would change it in editor config:
https://github.com/FriendsOfSymfony1/symfony1/blob/master/.editorconfig#L5

thePanz:
> @thePanz If you change the indent size in files (which is very fine by me) you would change it in editor config: https://github.com/FriendsOfSymfony1/symfony1/blob/master/.editorconfig#L5

hum, that should be extended for different file extensions, right?

btw: I haven't got time to dig into the failures of the tests... any help there? :blush:

connorhu:
> > @thePanz If you change the indent size in files (which is very fine by me) you would change it in editor config: https://github.com/FriendsOfSymfony1/symfony1/blob/master/.editorconfig#L5
> 
> hum, that should be extended for different file extensions, right?
> 
> btw: I haven't got time to dig into the failures of the tests... any help there? 😊

There is the fix:
https://github.com/FriendsOfSymfony1/symfony1/compare/feature/add-php-cs-fixer...connorhu:symfony1:feature/add-php-cs-fixer

My browser is died under the huge change set so I can't create a review to fix.

thePanz:
> > > @thePanz If you change the indent size in files (which is very fine by me) you would change it in editor config: https://github.com/FriendsOfSymfony1/symfony1/blob/master/.editorconfig#L5
> 
> There is the fix: [feature/add-php-cs-fixer...connorhu:symfony1:feature/add-php-cs-fixer](https://github.com/FriendsOfSymfony1/symfony1/compare/feature/add-php-cs-fixer...connorhu:symfony1:feature/add-php-cs-fixer)

Kosonom @connorhu !
Added the changes, and updated the .editorconfig setting to use 4 spaces

PR #290, opened 2023-09-18
fix: avoid creating dynamic property $options

vimar:
Dynamic properties are deprecated since PHP8.2

### DIFF
--- a/lib/view/sfViewCacheManager.class.php
+++ b/lib/view/sfViewCacheManager.class.php
@@ -21,6 +21,8 @@
  */
 class sfViewCacheManager
 {
+    public $options = array();
+
     protected $cache;
     protected $cacheConfig = array();
     protected $context;


thirsch (approved):

lib/view/sfViewCacheManager.class.php
```diff
@@ -29,7 +29,8 @@ class sfViewCacheManager
     protected $routing;
     protected $request;
     protected $loaded = array();
-
+    protected $options = array();
```
alquerci, line 32: Dynamic property are public.

So It is a BC break to reduce its accessibility scope.

I recommend making it `public`.

alquerci (changes requested):

lib/view/sfViewCacheManager.class.php
vimar: Thanks for your feedback I have just made the correction.

thirsch: @alquerci Ups, just to let you know: We have done many of those changes for 8.1 and 8.2 compat in the past.

alquerci (approved):

mentalstring (approved):

JohannesTyra (approved):
Testet and checked.

vimar:
any idea when this PR will be merged into main branch? is @thePanz the only user allowed to review and merge it?

thePanz (approved):

thePanz:
Thanks for poking me @vimar , approved the PR, waiting for the CI to run!

JohannesTyra:
@thePanz Thx. Should we create a new realease? Or are other commits also waiting for merging into master?

thePanz:
I have more changes in the pipelines, can not guarantee when they will land tho. 
I guess you can fix this commit on your SF 1.x project, and composer will let you know when the new release will appear @JohannesTyra

PR #292, opened 2023-11-16
Update(github) Update github workflows

thePanz:
- [x] include cache for composer installs
- [x] fix php-cs-fixer warnings


thePanz:
@thirsch @mentalstring OK to merge this?

my plan is to soon have php v7.4 as the min version, thus
- have short array syntax as for doctrine1
- later move to php v8.1 as min requirement

WDYT?

thirsch (approved):
LGTM, as stated in the ticket about upgrading the php version, I'm fine with a new min version as everyone who needs an older version can stick to an older version of this repo.

I'd still not go for version 2, but 1.6 - composer wouldn't install it on platforms not matching the new requirement and for manual updaters, we should include a prominent "don't do it" somewhere in the readme.md

thePanz:
> LGTM, as stated in the ticket about upgrading the php version, I'm fine with a new min version as everyone who needs an older version can stick to an older version of this repo.
Yes, that's exactly the plan! :+1: 
As the efforts on this project are *solely* to make it working on newer versions of PHP without crashes or deprecations!

> I'd still not go for version 2, but 1.6 - composer wouldn't install it on platforms not matching the new requirement and for manual updaters, we should include a prominent "don't do it" somewhere in the readme.md

The readme already has a "do not use this repo" line IIRC, and yes I would keep the versioning to 1.x.y , or even 1.5.x (to not go to v1.6 at all)

If you approve this PR I will merge it :)

thirsch (approved):

thirsch:
> If you approve this PR I will merge it :)

Already done, but looks like my permissions are not sufficient ;)

thirsch (approved):

PR #293, opened 2023-11-19
Restore support of PHP <7.1

alquerci:
A follow-up #283


thePanz:
@alquerci should we close this as we moved to v7.4 as min PHP version?

PR #294, opened 2023-12-15
php8.3 in the party

connorhu:
If php8.3 is already released, let's do a test with this version.


connorhu:
I am not sure about what cause the single_space_around_construct warnings. This rule introduced at v3.16.0. This is not php version related problem.

thePanz:
It could be that the last time php-CS-Fixer was run, it was in a previous version, thus that check not "default".

Let's fix it in this PR, wdyt?

thirsch:
What exactly is php-cs-fixer criticizing here? To be honest, I don't see any issues in line #1 of the mentioned files?

alquerci:
include is a construct

[An exemple](https://github.com/FriendsOfSymfony1/symfony1/blob/f7972ef33adc7febee6d41406802f32d0fbd356a/test/functional/prodTest.php#L13)

[The doc](https://cs.symfony.com/doc/rules/language_construct/single_space_around_construct.html)

thirsch:
> `include` is a construct
> 
> [An exemple](https://github.com/FriendsOfSymfony1/symfony1/blob/f7972ef33adc7febee6d41406802f32d0fbd356a/test/functional/prodTest.php#L13)
> 
> [The doc](https://cs.symfony.com/doc/rules/language_construct/single_space_around_construct.html)

Thanks, in the end it's not the opening php tag marked in line one.

For me, it would be ok to fix it in this pr. Wdyt, @connorhu?

connorhu:
and it's done.
I removed the brackets. I prefer includes to look different from functions.

thirsch (approved):

alquerci (approved):

thePanz (approved):

thePanz:
Thanks @connorhu 
merged!

Issue #286, opened 2023-02-19
[phpunit] port the test system to phpunit


connorhu:
This is an issue to track the progression.

Branch: https://github.com/connorhu/symfony1/tree/feature/phpunit
Target php version: php7.4

Components:
- [x] action
- [x] addon
- [x] autoload
- [x] cache
- [x] command
- [ ] config
- [ ] controller
- [ ] database
- [ ] debug
- [ ] escaper 
- [ ] event
- [ ] exception
- [ ] filter
- [ ] fixtures
- [ ] form
- [ ] generator
- [ ] helper
- [ ] i18n
- [ ] log
- [ ] mailer
- [ ] plugin
- [ ] request
- [ ] response
- [ ] routing
- [ ] service
- [ ] storage
- [ ] task
- [ ] test
- [ ] user
- [ ] util
- [ ] validator
- [ ] view
- [ ] widget
- [ ] yaml

connorhu:
https://github.com/connorhu/symfony1/blob/feature/phpunit/tests/controller/sfWebControllerTest.php#L37-L51
sfWebControllerTest now fail because of type error.

```
       yield ['module/action?id=12', array(
            '',
            array(
                'module' => 'module',
                'action' => 'action',
                'id'     => 12,
            ),
        )];
```

:
```
--- Expected
+++ Actual
@@ @@
     1 => Array &1 (
         'module' => 'module'
         'action' => 'action'
-        'id' => 12
+        'id' => '12'
     )
 )

```
 
If we look at the test writer intent, then all numeric arguments must be converted to real numbers.

connorhu:
A copy from issue #285 

PHPUnit brings up the errors nicely. The first assert fails because of an `acceptParameter` method bug.

```
        $option = new sfCommandOption('foo', null, sfCommandOption::IS_ARRAY);
        $this->assertSame(true, $option->acceptParameter());

        $option = new sfCommandOption('foo', null, sfCommandOption::PARAMETER_OPTIONAL);
        $this->assertSame(true, $option->acceptParameter());

        $option = new sfCommandOption('foo', null, sfCommandOption::PARAMETER_REQUIRED);
        $this->assertSame(true, $option->acceptParameter());

        $option = new sfCommandOption('foo', null, sfCommandOption::PARAMETER_NONE);
        $this->assertSame(false, $option->acceptParameter());
```

Description of `acceptParameter` describes how it should work, but the implementation doesn't do that:

[...]

PR #270, opened 2022-11-29
[WIP] Fix build of php53 docker-compose server

alquerci:

PR #295, opened 2023-12-16
port tests to use phpunit

connorhu:
see #286 

WIP


phpunit.xml.dist
```diff
@@ -0,0 +1,31 @@
...
+         failOnWarning="true"
+         stderr="true">
+    <php>
+        <ini name="error_reporting" value="32767" />
+        <ini name="intl.default_locale" value="en" />
+        <ini name="intl.error_level" value="0" />
+        <ini name="memory_limit" value="-1" />
+        <ini name="apc.enable_cli" value="1" />
+
+        <env name="MEMCACHED_HOST" value="localhost" />
```
alquerci, line 16: I suggest configuring this environment variable on `.github/workflows/continuous-integration.yml`.

It will be hard to change its value on local environment, for example with docker.

connorhu: On local environment you can override the dist file with phpunit.xml. The main goal is to always run every test without skipping.

alquerci: `MEMCACHED_HOST` change depending on the infrastructure, but not depending on test things.

This variable should be configured on `docker-compose.yaml` or on `.env`.

To achieve that goal, do then, do not skip it on the test but make it fail.

connorhu: Oh i see. I missed your point. I guess we have to refactor our `docker-compose.yaml` later. The `docker-compose.yaml` is only for testing, but 1) we should make a test profile 2) alongside an "official" docker compose for dev/prod profile.

alquerci: I like the idea of an "official" docker compose for dev/prod profile.

The current `symfony/symfony` repository does not provide an "official" docker compose.
Maybe it will go on another repository like https://github.com/dunglas/symfony-docker.

test/bin/test
```diff
@@ -12,20 +12,17 @@
 # Configuration
```
alquerci, line 12: Changes done here are in conflicts with those done there.

https://github.com/FriendsOfSymfony1/symfony1/pull/270/files#diff-1d56fbd8b67d0ff6b08b293bba30e27816f34e8815d901685e133972646828a3

thePanz:
@connorhu would you extract the removal of the Version tag into its own PR?
This way we can have a more focused (and reduced) changeset :+1:

connorhu:
> @connorhu would you extract the removal of the Version tag into its own PR? This way we can have a more focused (and reduced) changeset 👍

Sorry, I don't know how it got there.

connorhu:
`php-cs-fixer 403: rate limit exceeded` ☹️

connorhu:
I studied some how-tos and changed my mind. For phpunit testing we will also prefer phar based installs. The composer/script section shows which version is preferred. This can be expanded later if we want to support higher phpunit versions. For now I don't see the sense in this, version 9 will be fine.

thePanz:
> I studied some how-tos and changed my mind. For phpunit testing we will also prefer phar based installs. The composer/script section shows which version is preferred. This can be expanded later if we want to support higher phpunit versions. For now I don't see the sense in this, version 9 will be fine.

I would suggest to use phive to install the phar files. We could just ship the '.phive/phive.xml' file in the repo, where versioning is done there.

Drawback: the phive will run the installation without checking the currently installed PHP version, we would need to tale some extra care of the versioning, or have 2 phive.xml for php7 and php8

Wdyt?

connorhu:
> 
> I would suggest to use phive to install the phar files. We could just ship the '.phive/phive.xml' file in the repo, where versioning is done there.
> 
> Drawback: the phive will run the installation without checking the currently installed PHP version, we would need to tale some extra care of the versioning, or have 2 phive.xml for php7 and php8
> 
> Wdyt?

I'm going to dig into this and see what we can do here with phive.

connorhu:
I have some concerns about this tool. It can do some things but it's a bit awkward to use.
- phars xml undocumented, no xsd
- writes the installed version into the config and you can't make it not do that
- you need a phars.dist.xml file which has no exact version. for us it's enough to say that we need phpunit ^9.6, the detailed version is not interesting
- you can't set an alternative configuration file for it

We can live with these gaps, but it's a bit meh.

PR #302, opened 2024-01-01
[WIP] fix(test): fail on php errors

alquerci:


alquerci:
@thePanz hey, I got it tests passes on GitHub action.

Let me know whether this PR has any chance to be merged.

lib/mailer/sfMailerSwiftMessage.php
```diff
@@ -0,0 +1,17 @@
...
+ * This file is part of the symfony package.
+ * (c) Fabien Potencier <fabien.potencier@symfony-project.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+class_exists('Swift');
+
+if (version_compare(Swift::VERSION, '6.0.0') >= 0) {
```
thirsch, line 13: This approach is a bit like https://github.com/FriendsOfSymfony1/symfony1/pull/237/files but way more elegant. 

Question is, if we are able to support swiftmailer 6 including the doctrine spools, do we really need to support swiftmailer 5?

lib/config/autoload/swift.php
thirsch, line 1: I do like this approach to support the files being loaded by composer.

test/bootstrap/functional.php
```diff
@@ -8,6 +8,10 @@
  * file that was distributed with this source code.
  */
 
+if (!defined('SF_TEST_WITHOUT_COMPOSER')) {
```
thirsch, line 11: @thePanz and @connorhu this will bring support for composer autoloading in the tests and a way towards getting rid of the submodules.

thirsch (approved):

lib/mailer/sfMailerSwiftMessage.php
alquerci: Good question, I guess not.

But it depends, on the ability of applications using swiftmailer 5 to migrate to swiftmailer 6.

alquerci: This opens a path to add support for `symfony/mailer`.

PR #304, opened 2024-01-03
Fix setrawcookie() deprecation on null parameter

thePanz:


lib/response/sfWebResponse.class.php
```diff
@@ -359,7 +359,8 @@ public function sendHttpHeaders()
         // cookies
         foreach ($this->cookies as $cookie) {
             $expire = isset($cookie['expire']) ? $cookie['expire'] : 0;
-            setrawcookie($cookie['name'], $cookie['value'], $expire, $cookie['path'], $cookie['domain'], $cookie['secure'], $cookie['httpOnly']);
+            $domain = isset($cookie['domain']) ? $cookie['domain'] : '';
+            setrawcookie($cookie['name'], $cookie['value'], $expire, $cookie['path'], $domain, $cookie['secure'], $cookie['httpOnly']);
```
thirsch, line 363: Are the other keys are always present?

thirsch (approved):

lib/response/sfWebResponse.class.php
thePanz: Not sure, hope other users will report issues for other cases :)
Ok to merge? wdyt?

thirsch: Sure, it's already an improvement :)

PR #308, opened 2024-01-13
minor fix: remove fourth argument

connorhu:
There is no fourth argument. Nice catch phpstan!

### DIFF
--- a/lib/config/sfCacheConfigHandler.class.php
+++ b/lib/config/sfCacheConfigHandler.class.php
@@ -92,7 +92,7 @@ class sfCacheConfigHandler extends sfYamlConfigHandler
         $lifeTime = !$enabled ? '0' : $this->getConfigValue('lifetime', $actionName, '0');
 
         // client_lifetime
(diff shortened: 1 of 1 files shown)


alquerci (approved):

thirsch (approved):

mentalstring (approved):

thePanz (approved):

thePanz:
Thanks @connorhu!

Do we have a phpstan integration planned? that would be great to have it in the pipeline too
IIRC the last time I tried to enable it, it crashed with some methods overrides where the return type changed from a value to an array

connorhu:
@thePanz At this time we have 800+ errors at level 9 (Which seems a lot, but in a much smaller project I last found 3000). I can see the sense of making a distribution stan config, but running it in a workspace is yet too early. I plan to run it locally sometimes and fix any bugs that arise.

connorhu:
@mentalstring https://github.com/FriendsOfSymfony1/symfony1/compare/master...connorhu:symfony1:feature/stanconfig It needs work because it stops after 50 internal errors. Generated classes are missing.

connorhu:
oh. 24658 errors if i setup properly. nice.

PR #309, opened 2024-01-14
fix(phpdoc) remove html code from throws phpdoc. phpstan don't like this

connorhu:
simple fix

### DIFF
--- a/lib/cache/sfCache.class.php
+++ b/lib/cache/sfCache.class.php
@@ -50,7 +50,7 @@ abstract class sfCache
      *
      * * lifetime (optional): The default life time (default value: 86400)
      *
-     * @throws <b>sfInitializationException</b> If an error occurs while initializing this sfCache instance
+     * @throws sfInitializationException If an error occurs while initializing this sfCache instance
      */
     public function initialize($options = array())
     {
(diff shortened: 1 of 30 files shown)


lib/cache/sfCache.class.php
```diff
@@ -50,7 +50,7 @@ public function __construct($options = array())
      *
      * * lifetime (optional): The default life time (default value: 86400)
      *
-     * @throws <b>sfInitializationException</b> If an error occurs while initializing this sfCache instance
+     * @throws sfInitializationException</b> If an error occurs while initializing this sfCache instance
```
thirsch, line 53: The closing tag should also be removed? I didn't create a suggestion, as it has to be removed in all changed files.

thirsch (changes requested):

lib/cache/sfCache.class.php
thePanz: @connorhu please remove the `</b>` tag too :+1: 

Did you check if we have HTML tags elsewhere in the docblocks?

connorhu: What the hell. How did I miss them?

thirsch (approved):

mentalstring (approved):

PR #312, opened 2024-01-18
add(QA) phpstan dist config

connorhu:


thePanz:
@connorhu I added something like this to exclude files not relevant for phpstan

```
parameters:
    excludePaths:
        analyse:
            - lib/vendor/swiftmailer/
            - lib/task/symfony/lime_symfony.php
            - lib/task/symfony/sfSymfonyTestTask.class.php
            - lib/task/generator/skeleton/
            # Templates for errors
            - lib/exception/data/
            # Templates for classes
            - lib/plugins/sfDoctrinePlugin/data/generator/
            - lib/plugins/sfDoctrinePlugin/lib/vendor/doctrine/
            # Templates for application
            - lib/plugins/sfDoctrinePlugin/test/functional/fixtures/apps/
```

connorhu:
@thePanz Thank you! Updated the config. Should we do the same as doctrine1? (php7.4 and php8.3 simulatan tests)

.phpstan/bootstrap.php
```diff
@@ -0,0 +1,13 @@
...
+
+class ProjectConfiguration extends sfProjectConfiguration
+{
+}
+
+class BaseForm extends sfForm
+{
+}
+
+include 'PEAR/REST/10.php';
```
thePanz, line 11: :see_no_evil:  why do we need PEAR?! :see_no_evil: 

Should we completely drop it? @thirsch IIRC you mentioned something like this?

composer.json
```diff
@@ -8,7 +8,8 @@
         "friendsofsymfony1/swiftmailer": "^5.4.13 || ^6.2.5"
     },
     "require-dev": {
-        "psr/log": "*"
+        "psr/log": "*",
+        "friendsofsymfony1/doctrine1": "*"
```
thePanz, line 12: should we define the min-versions here?

should swiftmailer be added too?

.phpstan/bootstrap.php
connorhu: plugin uses this. as you can see in my plugin loading improvement issue #342 i would drop it too. if only because the default symfony channels don't work. refer to propel discussion

composer.json
connorhu: Swiftmailer is in required section.
What is the min version should be?

thePanz: > Swiftmailer is in required section

:see_no_evil: true :sweat_smile: 

> What is the min version should be?

I would pick the latest tagged one IMO

connorhu: The expert on submodule removal is @thirsch . Paging dr @thirsch 
https://github.com/FriendsOfSymfony1/symfony1/pull/302#discussion_r1506470471

PR #318, opened 2024-01-19
Fix construction of sfValidatorError with an array of an array

### LINKED ISSUE #317: sfValidatorDate throws sfValidatorError with array as $value and therefore not usable in the message
When using the Date Validator and providing a Date `2024-11-31` (only 30 Days in November), the Validator correctly throws a `sfValidatorError` but with the value as *array*. This happens multiple times in that file...

The exact position it happend is inside the `convertDateArrayToString` function where the value is an array but not yet converted to a string.
```php
// convert array to date string
if (is_array($value))
{
  $value = $this->convertDateArrayToString($value);
}
```

image

When trying to use %value% in the error message it won't be replaced, because `getArguments` of the class correctly checks for an array and returns nothing in the end. Therefore nothing of the error message will be replaced.

image

flohoss:
So in error messages the use of %year% %month% and %day% becomes possible

Closes #317

### DIFF
--- a/lib/validator/sfValidatorDate.class.php
+++ b/lib/validator/sfValidatorDate.class.php
@@ -154,7 +154,7 @@ class sfValidatorDate extends sfValidatorBase
         // all elements must be empty or a number
         foreach (array('year', 'month', 'day', 'hour', 'minute', 'second') as $key) {
             if (isset($value[$key]) && !ctype_digit((string) $value[$key]) && !empty($value[$key])) {
-                throw new sfValidatorError($this, 'invalid', array('value' => $value));
+                throw new sfValidatorError($this, 'invalid', $value);
             }
         }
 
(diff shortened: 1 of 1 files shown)


lib/validator/sfValidatorDate.class.php
```diff
@@ -156,7 +156,7 @@ protected function convertDateArrayToString($value)
         // all elements must be empty or a number
         foreach (array('year', 'month', 'day', 'hour', 'minute', 'second') as $key) {
             if (isset($value[$key]) && !ctype_digit((string) $value[$key]) && !empty($value[$key])) {
-                throw new sfValidatorError($this, 'invalid', array('value' => $value));
+                throw new sfValidatorError($this, 'invalid', $value);
```
alquerci, line 159: This will break `sfValidatorError::getValue()` that depends on `value` key.

alquerci (changes requested):

lib/validator/sfValidatorDate.class.php
thePanz: unfortunately that's not covered by the tests, as they're green anyway :(

alquerci: That's human mutation testing. :)

PR #320, opened 2024-02-05
PHP 8.3 deprecated errors fixed

miamibc:
Fixed error when running any task: 

```
Creating of dynamic property myTaskClassName::$version is deprecated
```

Fixed by adding these properties.
Also found same mistake in other parts of code... will be continued.

### DIFF
--- a/lib/plugins/sfDoctrinePlugin/lib/generator/sfDoctrineFormFilterGenerator.class.php
+++ b/lib/plugins/sfDoctrinePlugin/lib/generator/sfDoctrineFormFilterGenerator.class.php
@@ -19,6 +19,17 @@
  */
 class sfDoctrineFormFilterGenerator extends sfDoctrineFormGenerator
 {
+
+    /**
+     * @var Doctrine_Table
+     */
+    public $table;
+
+    /**
+     * @var string
+     */
+    public $modelName;
+
     /**
      * Initializes the current sfGenerator instance.
      *
--- a/lib/plugins/sfDoctrinePlugin/lib/generator/sfDoctrineFormFilterGenerator.class.php
+++ b/lib/plugins/sfDoctrinePlugin/lib/generator/sfDoctrineFormFilterGenerator.class.php
@@ -19,17 +19,6 @@
  */
 class sfDoctrineFormFilterGenerator extends sfDoctrineFormGenerator
 {
-
-    /**
-     * @var Doctrine_Table
-     */
-    public $table;
-
-    /**
-     * @var string
-     */
-    public $modelName;
-
     /**
      * Initializes the current sfGenerator instance.
      *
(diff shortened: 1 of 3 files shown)


lib/task/sfBaseTask.class.php
```diff
@@ -17,6 +17,13 @@
  */
 abstract class sfBaseTask extends sfCommandApplicationTask
 {
+
+    protected $version;
+    protected $namespace;
```
connorhu, line 22: namespace, name, briefDescription, detailedDescription fields are defined here:
https://github.com/FriendsOfSymfony1/symfony1/blob/6354fd37141c8fd40af563041674b5fe07ca99de/lib/task/sfTask.class.php#L18-L22

lib/task/sfBaseTask.class.php
```diff
@@ -17,6 +17,13 @@
  */
 abstract class sfBaseTask extends sfCommandApplicationTask
 {
+
+    protected $version;
```
connorhu, line 21: I don't see anyone using the version field. What's the reason you added this?

lib/plugins/sfDoctrinePlugin/lib/generator/sfDoctrineFormFilterGenerator.class.php
```diff
@@ -19,6 +19,17 @@
  */
 class sfDoctrineFormFilterGenerator extends sfDoctrineFormGenerator
 {
+
+    /**
+     * @var Doctrine_Table
+     */
+    public $table;
```
connorhu, line 26: You should delete them from here, because you have already added them in the `sfDoctrineFormGenerator` class.

connorhu (changes requested):

lib/task/sfBaseTask.class.php
miamibc: Oh yes. Sorry, looks like $version is something custom i added long time ago to some tasks. Removed, no deprecated messages anymore.

lib/task/sfBaseTask.class.php
miamibc: Oh yes. Sorry, looks like $version is something custom i added long time ago to some tasks. Removed, no deprecated messages anymore.

miamibc:
Done. Removed munencessary changes. 
Tested with commands `./symfony` and `./symfony doctrine:build --all-classes`, no errors in output.

thirsch (changes requested):
Thank you for your contribution, could you please clean up the commits of the pr to have only one commit with the desired change? Our goal is to keep the commit history as clean as possible.

connorhu:
@thirsch Don't necessarily need squash. We can do squash with PR adoption, same result.

connorhu:
@thePanz you noticed the wrong branch (unlike me). Thank you!

PR #331, opened 2024-02-27
Set min versions of swiftmailer to 5.4.13 and 6.2.5 and updated swiftmailer submodule

thirsch:


thirsch:
@thePanz I've reverted my approach of removing the submodules, as our own tests here in the repo heavily rely on them. Composer autoloading is not implemented for them. Maybe we can tackle this issue in another pr, independent of the min version constraints?

connorhu:
It would not be bad to get rid of the submodule system.

thePanz (approved):

thePanz:
> @thePanz I've reverted my approach of removing the submodules, as our own tests here in the repo heavily rely on them. Composer autoloading is not implemented for them. Maybe we can tackle this issue in another pr, independent of the min version constraints?

Yeah, fine for me!
Thanks for the PR

I will squash and  merge it

Issue #329, opened 2024-02-19
[github] improve cs fixer configuration


connorhu:
It should be smart about two things:
- it should only check files that have changed (my original PR knew this, but the ability was lost somewhere)
- it should filter which violations can stay based on the PR patch. In the case of a positive PR statistic, only the violation in the rows affected by the PR can remain:
![diff](https://github.com/FriendsOfSymfony1/symfony1/assets/1845575/8c329b37-19fd-4809-96c4-b813c0c12318)

PR #338, opened 2024-03-13
Update php cs fixer

thePanz:
- [x] Update to v3.51.0
- [x] Update to use short-array-syntax
- [x] Run on PHP v7.4 to avoid false edits on higher PHP releases
- [x] Trigger lint on CI only if relevant files have been changed


connorhu (approved):

thePanz:
@connorhu does this cover #329 ? Or just partially?

connorhu:
partially. "location of violation should be well marked" task is a very complex one. I'm trying to implement the fix in the php-cs-fixer source, but I'm far from a solution.

PR #339, opened 2024-03-13
Refactor doker setup

thePanz:


connorhu:
`Refactor doker setup` => `Refactor docker setup`

test/unit/validator/sfValidatorFileTest.php
```diff
@@ -90,7 +90,7 @@ public function getMimeTypesFromCategory($category)
 $v = new testValidatorFile();
 $t->is($v->guessFromFileBinary($tmpDir.'/test.txt'), 'text/plain', '->guessFromFileBinary() guesses the type of a given file');
 $t->is($v->guessFromFileBinary($tmpDir.'/foo.txt'), null, '->guessFromFileBinary() returns null if the file type is not guessable');
-$t->like($v->guessFromFileBinary('/bin/ls'), (PHP_OS != 'Darwin') ? '/^application\/x-(pie-executable|executable|sharedlib)$/' : '/^application/octet-stream$/', '->guessFromFileBinary() returns correct type if file is guessable');
+$t->is($v->guessFromFileBinary('/usr/bin/file'), 'application/x-pie-executable', '->guessFromFileBinary() returns correct type if file is guessable');
```
mentalstring, line 93: Won't this make the check pass only under some OS/distros? In other words, do we want to tests to only work under the docker setup?

thePanz: I guess so, this is to simplify the tests and the pipelines IMO.
Having a reliable docker setup would mean that we can rely on a single and common OS where tests are run.

Otherwise, to be consistent, we would need to have a pipeline testing that part for different OSes.

WDYT?

connorhu:
Please update the `README.md` file too if anything changes.

test/unit/validator/sfValidatorFileTest.php
mentalstring: It's definitely great that there's a consistent setup to run tests in, specially one that it can be used to easily run them against multiple php versions without having to maintain that mess locally – it's great for that and I'm all for it.

At the same time, it seems a pity that we opt to drop the possibility of easily running (by easy I mean having them pass) the whole test suite in different environments (be it macos, windows or some random linux distro) and adding the requirement to install/start/use docker to run the test suite.

Also, by making it less easy to run the tests in multiple contexts (e.g. by developers) we also lose some visibility into scenarios where the code&tests may be failing outside of the static docker setup which we may end up finding only when already running the code in some production server, where the set up is always going to be different than the docker one.

> Otherwise, to be consistent, we would need to have a pipeline testing that part for different OSes.

I don't think we need to have CI under different OS. That line was already updated a few times over the years to account for some OS specificities that different developers bumped into. Isn't it possible to tweak it to include whatever alpine does differently, and allow the test suite to continue to run (read pass) without docker?

mentalstring: PS: One way to at least avoid different distro layouts might be to test the function against `PHP_BINARY` which should always return an executable.

thePanz: @mentalstring I like your suggestion to use `PHP_BINARY`! nice trick! I applied the change and tests passed  :heavy_check_mark:

thePanz:
@connorhu @mentalstring I would need some help here: I did not setup any MySQL server and the tests are *all* green. Something is weird IMO

connorhu:
@thePanz Anno I started looking at the possibility of psr-4 migration and even with obvious problems the test was green with lime. Then I began to port the test system to phpunit because I found lime to be totally unreliable.
On topic: how can I help? Should I look at your branch?

test/unit/validator/sfValidatorFileTest.php
mentalstring: Actually, another idea to get this to run across different systems, hopefully definitely.

I don't see a good reason to be doing this test against a random executable file which may return different things in different systems – the `binary` in the name of the function refers to using the system `file` util, not that it only tests binary files.

Hence, instead of `PHP_BINARY`, test it against __FILE__ and expect `'text/x-php'` – from a quick test, this works both in linux and macos (`guessFromFileBinary()` doesn't work in windows).

mentalstring:
> @connorhu @mentalstring I would need some help here: I did not setup any MySQL server and the tests are _all_ green. Something is weird IMO

Not sure if this is what you are referring to, but I think only `unit/storage/sfMySQLStorageTest.php` would use MySQL and [it skips the test](https://github.com/FriendsOfSymfony1/symfony1/blob/50c946b4415691df3d2ece73c63364777f70cd2e/test/unit/storage/sfMySQLStorageTest.php#L18C1-L22) if the mysql extension is not installed. `unit/storage/sfMySQLiStorageTest.php` does the same if sqlite is not installed.

Currently, running tests with the `symfony:test` hides skipped tests behind a 'ok' which is how sfAPCCacheTest wasn't running on the CI for a long while due to test being skipped.

It also hides a bunch notices/warnings (e.g. try `php unit/validator/sfValidatorSchemaTest.php`). I think there is some work on https://github.com/FriendsOfSymfony1/symfony1/pull/302 that maybe helps with this, but I didn't look closely.

mentalstring:
Speaking of tests and refactoring the docker setup, I think `test/bin/test` is no longer needed?

PR #279, opened 2023-02-07
Adapt to Swiftmailer 5 and Swiftmailer 6 version of class Swift_ConfigurableSpool

thirsch:
@thePanz as promised, here is a possible solution of the swiftmailer issue. What do you think?

Btw. as the swiftmailer package is discontinued but needs a small adjustment for php 8.2: Should we clone the official repo in fos1 as well or try to adapt to symfony/mailer?

PR #346, opened 2024-03-21
[DoctrinePlugin] fix Swift_DoctrineSpool::queueMessage argument compatibility error

connorhu:
phpstan reported error


connorhu:
Not so nice the green status with an php error. :(
Our testing system is not good enough yet.

connorhu:
> IIRC this depends on the version of SwiftMailer installed, or I am wrong?

If this is true, we are not compatible with the latest version.

alquerci:
@thePanz https://github.com/FriendsOfSymfony1/symfony1/pull/302/files#diff-5d71b552c7a0ef74e4a9d348d29f677871db2cd0b8098308d94932d87b030dbd

thirsch:
> > IIRC this depends on the version of SwiftMailer installed, or I am wrong?
> 
> If this is true, we are not compatible with the latest version.

Unfortunately, that's true. But changing it with this pr would break support for 5.x - Either we do drop 5.x or we should use a patch like the one from @alquerci. Another approach can be found in https://github.com/FriendsOfSymfony1/symfony1/pull/279.

connorhu:
To summarize: we support two versions of a library that is no longer being developed, another one should be used, but the right direction would be that the framework does not require any mail-related library. Nice.

Issue #348, opened 2024-03-21
[QA] fix "return statement missing" errors


connorhu:
Errors like: "Function|Method ... should return ... but return statement is missing"

We should create PR by components:

- [ ] cache
- [ ] config
- [ ] debug
- [ ] exception
- [ ] filter
- [ ] helper
- [ ] i18n
- [ ] DoctrinePlugin
- [ ] routing
- [ ] service
- [x] task
- [x] util
- [x] validator
- [x] view
- [x] widget

<details>
  <summary>Detailed report</summary>

   ```
	'message' => '#^Method sfAPCCache\\:\\:clean\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfAPCCache.class.php',
--
	'message' => '#^Method sfAPCCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfAPCCache.class.php',
--
	'message' => '#^Method sfAPCuCache\\:\\:clean\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfAPCuCache.class.php',
--
	'message' => '#^Method sfAPCuCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfAPCuCache.class.php',
--
	'message' => '#^Method sfEAcceleratorCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfEAcceleratorCache.class.php',
--
	'message' => '#^Method sfFileCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfFileCache.class.php',
--
	'message' => '#^Method sfMemcacheCache\\:\\:clean\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfMemcacheCache.class.php',
--
	'message' => '#^Method sfMemcacheCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
	'path' => __DIR__ . '/lib/cache/sfMemcacheCache.class.php',
--
	'message' => '#^Method sfXCacheCache\\:\\:removePattern\\(\\) should return bool but return statement is missing\\.$#',
[...]

PR #350, opened 2024-03-21
[view] fix return statement missing stan error

connorhu:
issue: #348

### DIFF
--- a/lib/view/sfViewCacheManager.class.php
+++ b/lib/view/sfViewCacheManager.class.php
@@ -443,7 +443,7 @@ class sfViewCacheManager
             return $this->cache->remove($cacheKey);
         }
 
-        return false;
+        return true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/view/sfViewCacheManager.class.php
```diff
@@ -442,6 +442,8 @@ public function remove($internalUri, $hostName = '', $vary = '', $contextualPref
         if ($this->cache->has($cacheKey)) {
             return $this->cache->remove($cacheKey);
         }
+
+        return false;
```
thePanz, line 446: I would return `true` here, as if the key is not there, it is like is has been removed successfully.

Wdyt?

connorhu: true. i mean you are right.

lib/view/sfViewCacheManager.class.php
```diff
@@ -442,6 +442,8 @@ public function remove($internalUri, $hostName = '', $vary = '', $contextualPref
         if ($this->cache->has($cacheKey)) {
             return $this->cache->remove($cacheKey);
         }
+
+        return true;
```
thePanz, line 446:
```suggestion
        if (!$this->cache->has($cacheKey)) {
            return true;
        }

        return $this->cache->remove($cacheKey);
```

thePanz (approved):

PR #351, opened 2024-03-21
[util] fix return statement missing stan error

connorhu:
issue #348

### DIFF
--- a/lib/util/sfContext.class.php
+++ b/lib/util/sfContext.class.php
@@ -343,8 +343,7 @@ class sfContext implements ArrayAccess
     /**
      * Retrieve the module directory for this context.
      *
-     * @return string|null an absolute filesystem path to the directory of the
-     *                     currently executing module, if one is set, otherwise null
+     * @return string|null an absolute filesystem path to the directory of the currently executing module if one is set, null otherwise
      */
     public function getModuleDirectory()
     {
--- a/lib/util/sfContext.class.php
+++ b/lib/util/sfContext.class.php
@@ -359,8 +359,7 @@ class sfContext implements ArrayAccess
     /**
      * Retrieve the module name for this context.
      *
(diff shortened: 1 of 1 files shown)


lib/util/sfContext.class.php
```diff
@@ -341,8 +343,8 @@ public function getDatabaseManager()
     /**
      * Retrieve the module directory for this context.
      *
-     * @return string an absolute filesystem path to the directory of the
-     *                currently executing module, if one is set, otherwise null
+     * @return string|null an absolute filesystem path to the directory of the
+     *                     currently executing module, if one is set, otherwise null
```
thePanz, line 347:
```suggestion
     * @return string|null an absolute filesystem path to the directory of the currently executing module, if one is set, otherwise null
```

lib/util/sfContext.class.php
```diff
@@ -351,13 +353,15 @@ public function getModuleDirectory()
...
+        return null;
     }
 
     /**
      * Retrieve the module name for this context.
      *
-     * @return string the currently executing module name, if one is set,
-     *                otherwise null
+     * @return string|null the currently executing module name, if one is set,
+     *                     otherwise null
```
thePanz, line 364:
```suggestion
     * @return string|null the currently executing module name if one is set, null otherwise
```

lib/util/sfContext.class.php
```diff
@@ -235,8 +235,8 @@ public function getEventDispatcher()
     /**
      * Retrieve the action name for this context.
      *
-     * @return string the currently executing action name, if one is set,
-     *                otherwise null
+     * @return string|null the currently executing action name, if one is set,
+     *                     otherwise null
```
thePanz, line 239:
```suggestion
     * @return string|null the currently executing action name if one is set, null otherwise
```

PR #353, opened 2024-03-21
[task] fix return statement missing stan errors

connorhu:
issue #348 

Here I'm not sure whether not to introduce constants like symfony/console does:
https://github.com/symfony/symfony/blob/759b6e1126ed7a6bca6fe8f685bf47254aa82496/src/Symfony/Component/Console/Command/Command.php#L38-L40

### DIFF
--- a/lib/task/test/sfTestFunctionalTask.class.php
+++ b/lib/task/test/sfTestFunctionalTask.class.php
@@ -108,5 +108,7 @@ EOF;
 
             return $ret;
         }
+
+        return 0;
     }
 }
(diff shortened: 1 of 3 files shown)


thePanz:
> issue #348
> 
> Here I'm not sure whether not to introduce constants like symfony/console does: https://github.com/symfony/symfony/blob/759b6e1126ed7a6bca6fe8f685bf47254aa82496/src/Symfony/Component/Console/Command/Command.php#L38-L40

Maybe not now, we could do that with a rectorphp rule later on IMO

connorhu:
@thePanz ok! I'll resolve the conflict and do a rebase.

thePanz (approved):

Issue #354, opened 2024-03-22
update SYMFONY_VERSION constant accordingly


verfriemelt-dot-org:
heyho, just a minor nitpick, can you ensure to update the version defined here: 

https://github.com/FriendsOfSymfony1/symfony1/blob/master/lib/autoload/sfCoreAutoload.class.php#L12

:) thx for you effort with this legacy codebase!

connorhu:
Thank you for the reminder! We will add this step to the release process. Am I right @thePanz  ? :) (Just for the sake of science: what if I make a GHA process that does the release process on tagging. Including version number setting? I wonder what that would look like.).

thePanz:
> Thank you for the reminder! We will add this step to the release process. Am I right @thePanz ? :) (Just for the sake of science: what if I make a GHA process that does the release process on tagging. Including version number setting? I wonder what that would look like.).

Having an automated tagging would be great, or a pipeline making sure that the tagging is correct.
Never did that before tho

connorhu:
@thePanz Sebastian has just done one of his projects. What do you think about such a solution? https://github.com/sebastianbergmann/foal/commit/b0e3537cfc444803ae3d977510172684db24b697

j-sharpmoon:
Hi all,  I agree this is very minor. Yet after updating today, this false version output caused me to waste a lot of time mistakenly thinking that my git submodules were scrambled and somehow getting stuck at an earlier revision. 

Thanks to all of you who keep Symfony 1 working!

PR #355, opened 2024-03-22
update the SYMFONY_VERSION constant to the latest release number

verfriemelt-dot-org:
hi,

just a minor fix in context with #354 

cheers :v:

### DIFF
--- a/lib/autoload/sfCoreAutoload.class.php
+++ b/lib/autoload/sfCoreAutoload.class.php
@@ -9,7 +9,7 @@
  */
 
 // The current symfony version.
-define('SYMFONY_VERSION', '1.5.16-dev');
+define('SYMFONY_VERSION', '1.5.20-dev');
 
 /**
  * sfCoreAutoload class.


connorhu (approved):

thePanz:
@connorhu @verfriemelt-dot-org  it should be 1.5.20-dev, as 1.5.19 has been tagged few days ago :)

connorhu:
@thePanz Are we here already? Time flies (and version numbers)

verfriemelt-dot-org:
i updated it to 1.5.20 :) i think this was at some time done automatically via https://github.com/FriendsOfSymfony1/symfony1/blob/master/data/bin/release.php#L85
maybe a simple version file would be a simple solution.

something like this in the release pipeline:

```sh
$ git describe --tags > version
```

and just reading in that file from the lib-root instead of replacing that token in the php file.
with that we could just leave the version at `main` or `1.5.x-dev` :thinking:

thePanz:
> [...]with that we could just leave the version at `main` or `1.5.x-dev` 🤔

Having '1.5.x-dev' would be nice, we might also want to rename the 'main' branch to be '1.5.x' too, to simplify the management of -dev aliases (loke we have I for swiftmailer)

Wdyt?

thePanz (approved):

PR #358, opened 2024-03-23
[cs] added inactive header comment rule

connorhu:
This PR adds a header comment similar to the Symfony project. It is inactive because it generates a lot of noise and tasks when active. When touching a file it generates minimal noise and tasks when activated manually.


thePanz:
Should we enable it soon or later?
Would be cool to have a command to run that only on modified files on a PR

connorhu:
> Should we enable it soon or later?

Later. Sometime it duplicates a header comment so this generates a lot of unwanted noises and codes to fix.

Today I thought of such a solution:
```php
<?php

/** @var PhpCsFixer\Config $config */
$config = require_once __DIR__.'/.php-cs-fixer.dist.php';

$config->setRules($config->getRules() + [
    'header_comment' => [
        'header' => <<<'EOF'
This file is part of the Symfony1 package.

(c) Fabien Potencier <fabien@symfony.com>

For the full copyright and license information, please view the LICENSE
file that was distributed with this source code.
EOF,
    ],
]);

return $config;
```
It's a solution where risky, task-generating, not-everywhere things can be turned on manually:
`~/Applications/phpunit/php-cs-fixer-3.phar fix --config=.php-cs-fixer.header.php`

connorhu:
I close this for now. My other solution is enough for now.

PR #359, opened 2024-03-23
[action] improvement: use null coalescing operator

connorhu:
less code happier maintainer

### DIFF
--- a/lib/action/sfAction.class.php
+++ b/lib/action/sfAction.class.php
@@ -388,15 +388,7 @@ abstract class sfAction extends sfComponent
     {
         $actionName = strtolower($this->getActionName());
 
-        if (isset($this->security[$actionName][$name])) {
-            return $this->security[$actionName][$name];
-        }
-
-        if (isset($this->security['all'][$name])) {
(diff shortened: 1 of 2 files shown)


connorhu:
@thePanz 

This kind of code has a simplified version:
```php
        if (!isset(self::$instance)) {
            self::$instance = new sfAutoload();
        }

        return self::$instance;
```

```php
        return self::$instance ?? self::$instance = new sfAutoload();
```

WDYT?

lib/action/sfAction.class.php
```diff
@@ -290,7 +290,7 @@ public function getPartial($templateName, $vars = null)
     {
         $this->getContext()->getConfiguration()->loadHelpers('Partial');
 
-        $vars = null !== $vars ? $vars : $this->varHolder->getAll();
+        $vars ??= $this->varHolder->getAll();
```
thePanz, line 293: this is not supported by php v7.4 IIRC

connorhu: https://wiki.php.net/rfc/null_coalesce_equal_operator

PR #360, opened 2024-03-27
[misc] release process refactor

connorhu:


connorhu:
- I would separate the scripts I wrote from the data/bin library (which I don't like anyway). They are maintenance scripts. The place where I put them now is not quite right for them, I should find a better place for them.
- I minimized the use of 3rd party libraries. The `lime_output` class could be used for output coloring, but if we ever get to the point where phpunit takes the place of lime I'd rather not use it and use minimal tools.

PR #362, opened 2024-03-27
Optimize readability for `Label` @ `sfWidgetFormSchemaFormatter`

xerc:


lib/widget/sfWidgetFormSchemaFormatter.class.php
```diff
@@ -168,7 +168,7 @@ public function generateLabelName($name)
         $label = $this->widgetSchema->getLabel($name);
 
         if (!$label && false !== $label) {
-            $label = str_replace('_', ' ', ucfirst('_id' == substr($name, -3) ? substr($name, 0, -3) : $name));
+            $label = sfInflector::camelize('_id' == substr($name, -3) ? substr($name, 0, -3) : $name);
```
connorhu, line 171: This cause BC break:
```php

not ok 16 - ->generateLabelName() generates a label value from a label name
#     Failed test (./test/unit/widget/sfWidgetFormSchemaFormatterTest.php at line 123)
#            got: 'FirstName'
#       expected: 'First name'

```

connorhu (changes requested):

lib/widget/sfWidgetFormSchemaFormatter.class.php
xerc: yes ; what format do you/others prefere ? `FirstName` | `First Name` | `First name`

connorhu: "First name" As the test expects.

PR #366, opened 2024-04-04
[cs] introduce modern string functions

connorhu:
This PR is introduce the `str_starts_with` `str_ends_with` `str_contains` functions where is possible.
It improves code readability and also php8 or above cause an improvement in performance.

Mostly automated fixes are done using Rector and php-cs-fixer. Relevant rector rules:
- \Rector\Php80\Rector\Identical\StrEndsWithRector
- \Rector\Php80\Rector\NotIdentical\StrContainsRector
- \Rector\Php80\Rector\Identical\StrStartsWithRector

For php 7.4 `symfony/polyfill-php80` provides backwards compatibility.

Template for rector:
```php
<?php

use Rector\Config\RectorConfig;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__.'/lib',
        __DIR__.'/test',
    ]);
    $rectorConfig->skip([
        __DIR__.'/lib/*/skeleton/*',
    ]);
```

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -29,7 +29,9 @@ $config
         '@PhpCsFixer' => true,
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'short'],
+        'modernize_strpos' => true,
     ])
+    ->setRiskyAllowed(true)
     ->setCacheFile('.cache/php-cs-fixer.cache')
     ->setFinder($finder)
 ;
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,8 @@
     "license": "MIT",
     "require": {
         "php" : "^7.4 || ^8.1",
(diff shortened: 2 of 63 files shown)


connorhu:
A skeleton jumped out of the closet. The problem is that the current codebase doesn't recognize and use the composer autoloader, everyone has to load it themselves (see migration wiki).

alquerci (approved):

alquerci (commented):
I am not sure.

Indeed adding new dependency that requires the use of composer is a BC break.

As composer is not yet required.

connorhu:
> I am not sure.
> 
> Indeed adding new dependency that requires the use of composer is a BC break.
> 
> As composer is not yet required.

Yes, this is a valid point. This is indeed a BC break in the sense that there was no composer req. before (although we support it). I don't see this as a problem in the sense that composer is no cutting edge today and this change is more about bringing practices that have become standard.

Issue #341, opened 2024-03-20
generate:project command is broken


connorhu:
Although it is not recommended to create a new project based on symfony1, it may sometimes be necessary to build a symfony1 project from scratch.

For a completely empty project, the `project:generate` command does not work.

Error:
```
PHP Warning: require_once(/Doctrine/Core.php): Failed to open stream: No such file or directory in .../vendor/friendsofsymfony1/symfony1/lib/plugins/sfDoctrinePlugin/config/sfDoctrinePluginConfiguration.class.php on line 36

PHP Fatal error: Uncaught Error: Failed opening required '/Doctrine/Core.php' (include_path='.:/Users/connor/.phpbrew/php/php-8.3.0/lib/php/pear') in .../vendor/friendsofsymfony1/symfony1/lib/plugins/sfDoctrinePlugin/config/sfDoctrinePluginConfiguration.class.php:36
```

Code fragment causing the error:
```php
        if (!class_exists('Doctrine_Core')) {
            require_once sfConfig::get('sf_doctrine_dir', realpath(dirname(__FILE__).'/../lib/vendor/doctrine/lib')).'/Doctrine/Core.php';
            spl_autoload_register(['Doctrine_Core', 'autoload']);
        }
```

Reproduction steps:
```
mkdir folder
cd folder
composer require friendsofsymfony1/symfony1
vendor/bin/symfony generate:project gentest
```

We should check if the doctrine plugin can be detached from the project and moved to a separate repo.

mentalstring:
It seems to work with `--orm=none`, because it defaults to Doctrine.

Anyway, I agree that having it shipped with the framework doesn't quite make sense, although I'm partial as I only use Propel with symfony1. Having it in a separate package would also facilitate upgrading one separate from the other.

That said, I'm not sure how much work it be to do that. Perhaps a easier route on the short term could be to have it run out of the `plugins` directory as I think it originally used to.

connorhu:
@mentalstring It does not seem very complicated. My biggest dilemma is how to do the separation: actually, or in a subtree layout, as the symfony does.

PR #367, opened 2024-04-04
[command] allow to call symfony command from anywhere

### LINKED ISSUE #204: `symfony` as symlink (callable from anywhere)
```
// prevent symlink obfuscation
chdir(dirname($_SERVER['SCRIPT_FILENAME']));
```
https://github.com/FriendsOfSymfony1/symfony1/blob/master/data/bin/symfony#L12-L15

connorhu:
closes #204 

What this cause?
a) cwd is set properly => nothing
b) cwd is not set properly => somewhere something cause error

But it's the caller's responsibility to make sure cwd is set properly.
The symfony command sets cwd anyway:
https://github.com/FriendsOfSymfony1/symfony1/blob/0c9ba112fc2d7ed21796079fb66c5e592943c701/lib/task/generator/skeleton/project/symfony#L12

### DIFF
--- a/lib/task/sfBaseTask.class.php
+++ b/lib/task/sfBaseTask.class.php
@@ -108,8 +108,6 @@ abstract class sfBaseTask extends sfCommandApplicationTask
             return $event->getReturnValue();
         }
 
-        $this->checkProjectExists();
-
         $requiresApplication = $commandManager->getArgumentSet()->hasArgument('application') || $commandManager->getOptionSet()->hasOption('application');
         if (null === $this->configuration || ($requiresApplication && !$this->configuration instanceof sfApplicationConfiguration)) {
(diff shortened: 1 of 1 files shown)


alquerci (changes requested):
LGTM, but a BC break concern.

#### About `checkProjectExists()`

As a public method, we do not know the usage of this public method.
Changing its behaviour is a BC break.

Only removing its usage, will be enough.

I suggest keeping it untouched OR use `config/ProjectConfiguration.class.php` like
https://github.com/FriendsOfSymfony1/symfony1/blob/573d5790cb007ff8a4447a148497a5516673fcea/data/bin/symfony#L12-L15

connorhu:
This feature protects the developer during development from doing the inappropriate thing (e.g. calling cli.php in the wrong way). But on a complete system, this method is no longer relevant and as I wrote, the cwd is handled by the symfony script. There are already tools used during development which are not tested and do not work properly under certain circumstances (e.g. #341).

PR #371, opened 2024-04-09
[WIP] fix(test): exit code of lime test -- add proof of current behaviour

alquerci:


lib/vendor/lime/lime.php
```diff
@@ -1186,8 +1199,7 @@ public function process($files)
 EOF;
       file_put_contents($tmp_file, $tmp);
       ob_start();
-      // see http://trac.symfony-project.org/ticket/5437 for the explanation on the weird "cd" thing
```
connorhu, line 1189: **[PATCH] lime.php not working on windows when there are spaces in php interpreter path**

lime.php, line 541
```php
passthru(sprintf('%s "%s" 2>&1', $this->php_cli, $file), $return);
```

the command fails if the path to php interpreter contains spaces, this obviously prevent all test to run.

In order to work with executables in directories with spaces it should be
```php
passthru(sprintf('"%s" "%s" 2>&1', $this->php_cli, $file), $return);
```
but this fails too on windows: system/exec/passthru all fail when more than two arguments are quoted (it seems a php bug but I couldn't find any reference apart from some comments on the [system function page](https://www.php.net/manual/en/function.system.php))

In order to work on windows it should be written as
```php
passthru(sprintf('""%s" "%s"" 2>&1', $this->php_cli, $file), $return);
```
this works on windows but still fails on linux (note the double "double quotes": I need to quote the entire command not just the executable and the argument)

The only workaround I could find was something I found on the comments of the [system function](https://www.php.net/manual/en/function.system.php) that is adding a "nop" command call before the quoted command
```php
passthru(sprintf('cd & "%s" "%s" 2>&1', $this->php_cli, $file), $return);
```

test/unit/vendor/lime/fixtures/pass_with_one_parse_error.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+require_once __DIR__.'/../../../../bootstrap/unit.php';
+
+$test = new lime_test();
+
+parse_error
```
connorhu, line 7: This file should be excluded from everywhere that parses php files. At this moment php-cs-fixer, but later phpstan, rector etc.

test/unit/vendor/lime/lime_harnessTest.php
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    private function whenExecuteHarnessWithOneFileWillHaveStatusCodeAndOutput($name, $expectedOverallSucceed, $expectedOutput)
+    {
+        $this->info($name);
+
+        $harness = $this->makeHarness();
+
+        $harness->register([
+            __DIR__."/fixtures/$name.php",
```
connorhu, line 24:
```suggestion
            __DIR__.'/fixtures/$name.php',
```

test/unit/vendor/lime/lime_harnessTest.php
```diff
@@ -0,0 +1,261 @@
+<?php
+
+require_once __DIR__.'/../../../bootstrap/unit.php';
+require_once __DIR__.'/tools/TestCase.php';
+require_once __DIR__.'/tools/lime_no_colorizer.php';
```
connorhu, line 5: What is the filename case policy in this directory?

connorhu (changes requested):
Clear and understandable. Thank you!

test/unit/vendor/lime/lime_harnessTest.php
alquerci: The $name variable need to be interpreted.

Using single quotes change the behaviour.

test/unit/vendor/lime/lime_harnessTest.php
alquerci: Same as class name.

test/unit/vendor/lime/fixtures/pass_with_one_parse_error.php
alquerci: php-cs-fixer exclude all files that contain "vendor" on its path.

This means that the directory "test/unit/vendor" is ignored.

Fixing that, will requires a structural correction, that is out of scope of this PR.

@connorhu What can I do on the scope of this PR?

test/unit/vendor/lime/lime_harnessTest.php
alquerci: Here, I had think about passing the file path (concrete) or the name (intent).

I choose to pass the intent.

connorhu: Sorry my bad.

test/unit/vendor/lime/lime_harnessTest.php
```diff
@@ -0,0 +1,261 @@
...
+{
+    private function makeHarness(): lime_harness
+    {
+        $harness = new lime_harness();
+        $harness->output->colorizer = new lime_no_colorizer();
+
+        return $harness;
+    }
+
+    private function whenExecuteHarnessWithOneFileWillHaveStatusCodeAndOutput($name, $expectedOverallSucceed, $expectedOutput)
```
alquerci, line 17: The method name is not consistent with its content.

Need to choose between
- file or name
- status code or overall success

test/unit/vendor/lime/fixtures/pass_with_one_parse_error.php
connorhu: Indeed, you are right! For the other tools, it was just a reminder to myself.

test/unit/vendor/lime/lime_harnessTest.php
alquerci: The action here is to include directory "test/unit/vendor" on php-cs-fixer configuration, then execute the fixer.

test/unit/vendor/lime/fixtures/pass_with_one_parse_error.php
alquerci: 1. [configure php-cs-fixer](https://github.com/FriendsOfSymfony1/symfony1/pull/371/commits/4523808934ecc5841757b0d1c5dbbd684d54c16f)
2. [execute CS fixer](https://github.com/FriendsOfSymfony1/symfony1/pull/371/commits/54a52f8af7edfc7e41e0c590926b99e46f87e289)
3. [Allow having pretty method name for test cases](https://github.com/FriendsOfSymfony1/symfony1/pull/371/commits/7dccdd1e09cd02d5e6e3afbb7219f596b798bbfb)

test/unit/vendor/lime/lime_harnessTest.php
alquerci: Done in 3 steps

See https://github.com/FriendsOfSymfony1/symfony1/pull/371#discussion_r1559937756

Issue #379, opened 2024-12-13
Base test classes are missing from package builds


vworldat:
As already commented by @JohannesTyra at https://github.com/FriendsOfSymfony1/symfony1/commit/3f0adca225fbf9d14ea35bb2adcc6d12d4f42598#commitcomment-149925742 the change to include `test/` in `.gitattributes` not only messes with the autoloader, but also removes critical classes like `sfTestBaseTask` or `sfTester` from all git exports and thus composer packaging.

I think the line should be changed to `/test/` to only match the actual symfony1 tests that can be safely ignored.

thePanz:
@vworldat I checked on on of my SF1.5 projects, but was not able to reproduce the issue when running some of he SF1 actions.
Not sure if it is due to a different scenario, or the wrapping with SF5.4 I have in place.

Do you have a reproducible example we can test and replicate on the tests to avoid those breaking changes?

PR #380, opened 2024-12-13
Fix: inclusion of base test files in composer package (#379)

vworldat:
See #379

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,5 +8,5 @@
 docker-compose.yml     export-ignore
 phpstan.neon           export-ignore
 phpunit.xml            export-ignore
-test/                  export-ignore
+/test/                 export-ignore
 tests/                 export-ignore


thePanz:
Thanks @vworldat 
Should we anchor the other excluded files to the repository root too?

vworldat:
> Thanks @vworldat Should we anchor the other excluded files to the repository root too?

On a personal level, I usually prefer narrow definitions for directory paths because they are most likely to cause false positives. Looking at this project in particular, I'd say perhaps anchor the dirs, but keep the file names as they are. I see no reason to include any of the listed files in a package, unless there was a nested `.gitignore` that's only used to provide a mandatory empty directory.

thePanz:
Thanks @vworldat for your contribution... and PR!
feel free to open another PR to anchor them if you have time!

PR #386, opened 2025-04-17
minor fixes on sfFormField and sfSessionStorage

pedromellofh:
sfFormField lacks defaults in __construct and sfSessionRestore throws a warning when calls session_regenerate_id uppon an already  started


thePanz:
Thanks @alquerci for the pointer to SF 7.2 implementation.

@pedromellofh I would follow that example here, and NOT silence the call triggering the warning.

Would you adapt the code, by taking in the example from SF 7.2 linked by @alquerci ? thank you

PR #394, opened 2026-02-02
Fix I18n filterGet returning wrong culture after connection clear()

JohannesTyra:
## Problem

When changing the culture mid-request (e.g., to send an email in a user's preferred language), `filterGet()` returns translations in the wrong culture after calling `Doctrine_Connection::clear()`.

## Root Cause

PHP's `isset()` with chained array access like:
```php
isset($record['Translation'][$culture])
```
does **not** trigger `__get()` on the intermediate `$record['Translation']` access.

This means the Translation relation is never lazy-loaded, causing `isset()` to incorrectly return `false`. The method then falls back to `sf_default_culture` instead of the requested culture.

## Steps to Reproduce

```php
// 1. Load record and access translated field (triggers lazy load in culture A)
sfDoctrineRecord::setDefaultCulture('de');
$record = MyTable::getInstance()->find(1);
$title = $record->getTitle(); // Returns German

// 2. Clear the identity map
Doctrine_Manager::getInstance()->getCurrentConnection()->clear();

// 3. Change culture
sfDoctrineRecord::setDefaultCulture('fr');

// 4. Load fresh record and access translated field
$record2 = MyTable::getInstance()->find(1);
$title2 = $record2->getTitle();

// BUG: Returns German instead of French!
```

## Solution

Explicitly assign `$record['Translation']` to a variable before the `isset()` check. This triggers proper lazy loading of the Translation relation:

```php
$translation = $record['Translation'];  // Triggers lazy loading
if (isset($translation[$culture]) && '' != $translation[$culture][$name]) {
    return $translation[$culture][$name];
}
```

## Real-World Impact

This bug commonly affects applications that:
- Send emails in a user's preferred language during a request
- Process batch operations with different user cultures
- Switch cultures after clearing Doctrine's identity map

## Testing

The fix has been tested with multiple scenarios:
- No reload after culture change
- `refresh(true)` without `clear()`
- `refresh(true)` with `clear()`
- `find()` with `clear()`
- `find()` with `clear()` + `setDefaultCulture()`

All scenarios now correctly return translations in the target culture.

### DIFF
--- a/lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecordI18nFilter.class.php
+++ b/lib/plugins/sfDoctrinePlugin/lib/record/sfDoctrineRecordI18nFilter.class.php
@@ -46,12 +46,19 @@ class sfDoctrineRecordI18nFilter extends Doctrine_Record_Filter
     public function filterGet(Doctrine_Record $record, $name)
     {
         $culture = sfDoctrineRecord::getDefaultCulture();
-        if (isset($record['Translation'][$culture]) && '' != $record['Translation'][$culture][$name]) {
-            return $record['Translation'][$culture][$name];
+
+        // Access Translation relation explicitly to trigger lazy loading.
+        // PHP's isset() with chained array access like isset($record['Translation'][$culture])
+        // does not trigger __get() on the intermediate $record['Translation'] access,
+        // causing the check to incorrectly return false on first access after clear().
+        $translation = $record['Translation'];
+
+        if (isset($translation[$culture]) && '' != $translation[$culture][$name]) {
+            return $translation[$culture][$name];
         }
 
         $defaultCulture = sfConfig::get('sf_default_culture');
 
-        return $record['Translation'][$defaultCulture][$name];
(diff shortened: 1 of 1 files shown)


Tybaze:
Thanks for catching this, it may explain why I tend to always pass the culture explicitly when reading translated attributes.

Steps to reproduce with a simplified example:

```php

// 1. Switch to a culture which is not the default one, example with sf_default_culture = "en"
sfDoctrineRecord::setDefaultCulture('de');

// 2. Load fresh record and access translated field
$record = MyTable::getInstance()->find(1);
$title = $record->getTitle();

// BUG: $title is in English while this object has German text in the database
```

With your fix, we may incur additional SQL queries to load missing translations, whereas the current behavior does not. 

That said, I think the trade-off is acceptable: it aligns the behavior with the intent of the feature.

alquerci:
related to https://github.com/FriendsOfSymfony1/symfony1/pull/184/changes#r1180512181

thirsch (approved):

thirsch:
I'd go for this solution instead of the linked #184.

PR #395, opened 2026-05-13
⚠️ [BC-Break] Require ext-mbstring; replace utf8_encode in lime JUnit reporter

JohannesTyra:
## Summary

`utf8_encode()` has been deprecated since PHP 8.2. The lime JUnit reporter calls it on test names at `lib/vendor/lime/lime.php:106`:

```php
$testcase->setAttribute('name', utf8_encode($test['message']));
```

Switch to the documented migration path `mb_convert_encoding($s, 'UTF-8', 'ISO-8859-1')`.

PR #393 added PHP 8.5 support and did touch `lime.php`, but only swapped `(boolean)` for `(bool)` at line 169 — the `utf8_encode` call at line 106 was left in place.

## Verification

Under PHP 8.4.7:
```
Function utf8_encode() is deprecated since 8.2, visit the php.net documentation for various alternatives
```

## Test plan
- [ ] Existing test suite passes on the 8.4/8.5 CI matrix
- [ ] JUnit reporter output is byte-identical for ASCII test names (the common case) and unchanged Latin-1-encoded for any high-bit input

### DIFF
--- a/lib/vendor/lime/lime.php
+++ b/lib/vendor/lime/lime.php
@@ -103,7 +103,7 @@ class lime_test
       foreach ($result['tests'] as $test)
       {
         $testsuite->appendChild($testcase = $dom->createElement('testcase'));
-        $testcase->setAttribute('name', utf8_encode($test['message']));
+        $testcase->setAttribute('name', mb_convert_encoding($test['message'], 'UTF-8', 'ISO-8859-1'));
         $testcase->setAttribute('file', $test['file']);
         $testcase->setAttribute('line', $test['line']);
         $testcase->setAttribute('assertions', 1);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 xx/xx/xxxx: Version 1.5.xx
 --------------------------
+* ⚠️ [BC-Break] Require ext-mbstring; replace deprecated utf8_encode() in lime JUnit reporter by @JohannesTyra in #395
 
 26/12/2025: Version 1.5.24
 --------------------------
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
     "license": "MIT",
     "require": {
         "php" : "^7.4 || ^8.1",
+        "ext-mbstring": "*",
         "friendsofsymfony1/swiftmailer": "^5.4.13 || ^6.2.5"
     },
     "require-dev": {


lib/vendor/lime/lime.php
```diff
@@ -103,7 +103,7 @@ static public function to_xml($results = null)
       foreach ($result['tests'] as $test)
       {
         $testsuite->appendChild($testcase = $dom->createElement('testcase'));
-        $testcase->setAttribute('name', utf8_encode($test['message']));
+        $testcase->setAttribute('name', mb_convert_encoding($test['message'], 'UTF-8', 'ISO-8859-1'));
```
thePanz, line 106: should we require `mb_string` extension?

JohannesTyra:
Thanks @thePanz — good catch. Pushed 0b09d31: added `"ext-mbstring": "*"` to `composer.json` `require`. Now `mb_convert_encoding` is a safe call: if mbstring is missing, Composer fails at install time rather than the test reporter crashing at runtime.

Tybaze:
Adding a new dependency to such an old project is usually something we should be careful about.

That said, I think requiring `ext-mbstring` is acceptable here:

- `mbstring` adoption is very wide.
- The extension has existed for a very long time, so it is not a modern PHP-only requirement.
- It can be installed even on very old environments.
- Doctrine1, which is historically the main library associated with symfony1, has required `mbstring` for years without major complaints.
- The PHP documentation itself recommends replacing `utf8_encode()` / `utf8_decode()` with `mb_convert_encoding()` prior the alternatives such as `UConverter::transcode()` or `iconv`.

So my opinion is that requiring `ext-mbstring` is safer and cleaner than adding a custom fallback layer.

Open questions:

- Should we implement a fallback to `iconv` if `mbstring` is missing? Personally, I do not think so, because it adds complexity for a very limited benefit.
- Later, should we remove the tests that skip when `mbstring` is missing, since it would now be a project requirement?

Tybaze (approved):
OK to move to require mbstring

thePanz:
@JohannesTyra would you please add a line to our changelog about the added `ext-mbstring` requirement? Prefix it with `⚠️ [BC-Break] ` as done for previous changes

@Tybaze yeah, I am aligned with you, `mbstring` is quite a safe requirement, and avoids implementing a hack-fallback.

JohannesTyra:
Done — added in 2df4776 under the unreleased `1.5.xx` section. Thanks both for the reviews! @Tybaze @thePanz

thePanz (approved):

JohannesTyra:
Also bumped the PR title to include the `⚠️ [BC-Break]` marker, to match the changelog convention.

PR #396, opened 2026-05-29
PHP 8.4: register session save handler as a SessionHandlerInterface object

JohannesTyra:
## Problem

Under PHP 8.4, `sfDatabaseSessionStorage::initialize()` emits:

```
PHP Deprecated:  session_set_save_handler(): Providing individual callbacks instead of an object implementing SessionHandlerInterface is deprecated in .../lib/storage/sfDatabaseSessionStorage.class.php on line 66
```

PHP 8.4 deprecates the six-callback form of `session_set_save_handler()` in favour of passing an object implementing `SessionHandlerInterface`. This fires on every request served with a database session backend (e.g. `sfPDOSessionStorage`).

## Fix

`sfDatabaseSessionStorage` **cannot** implement `SessionHandlerInterface` directly: its parent `sfSessionStorage` already defines `read()`/`write()` for symfony's *attribute* storage (`$_SESSION[$key]`), which collide with the save-handler `read()`/`write()` and have different semantics.

So this adds a small `sfDatabaseSessionHandler` adapter that implements `SessionHandlerInterface` and delegates to the existing `sessionOpen` / `sessionClose` / `sessionRead` / `sessionWrite` / `sessionDestroy` / `sessionGC` methods. The handler is registered with `session_set_save_handler($handler, true)`.

`#[\ReturnTypeWillChange]` is used on the adapter methods because `SessionHandlerInterface` declares tentative return types and the legacy `sessionXxx()` methods return loosely-typed values that must be preserved unchanged (e.g. `sessionGC()` returns `bool`, not `int`).

The adapter is defined in the same file as `sfDatabaseSessionStorage` so it is always available wherever the core autoloader has loaded that class; it is only referenced from within `initialize()`.

## Behaviour

No behavioural change. Verified on PHP 8.4.21 by driving a full session lifecycle through the adapter (`open` → `read` → `gc` → `write` → `close`): correct delegation, **no deprecation, no TypeError**. Confirmed the legacy six-callback form still reproduces the deprecation and that omitting `#[\ReturnTypeWillChange]` produces tentative-return-type notices.

### DIFF
--- a/lib/storage/sfDatabaseSessionStorage.class.php
+++ b/lib/storage/sfDatabaseSessionStorage.class.php
@@ -189,6 +189,7 @@ abstract class sfDatabaseSessionStorage extends sfSessionStorage
 /**
  * Adapts an sfDatabaseSessionStorage to the SessionHandlerInterface required by
  * PHP 8.4's session_set_save_handler().
+ *
  * See https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#session_set_save_handler
  *
  * This must be a separate object rather than having sfDatabaseSessionStorage
--- a/lib/storage/sfDatabaseSessionStorage.class.php
+++ b/lib/storage/sfDatabaseSessionStorage.class.php
@@ -63,14 +63,9 @@ abstract class sfDatabaseSessionStorage extends sfSessionStorage
         }
 
         // use this object as the session handler
-        session_set_save_handler(
-            [$this, 'sessionOpen'],
-            [$this, 'sessionClose'],
-            [$this, 'sessionRead'],
-            [$this, 'sessionWrite'],
-            [$this, 'sessionDestroy'],
-            [$this, 'sessionGC']
-        );
+        // PHP 8.4: the 6-callback form of session_set_save_handler() is deprecated.
+        // Register a SessionHandlerInterface object instead (see adapter below).
+        session_set_save_handler(new sfDatabaseSessionHandler($this), true);
 
         // start our session
         session_start();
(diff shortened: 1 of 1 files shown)


lib/storage/sfDatabaseSessionStorage.class.php
```diff
@@ -190,3 +185,67 @@ public function regenerate($destroy = false)
...
+ * return types, and the legacy sessionXxx() methods return loosely-typed values
+ * (e.g. sessionGC() returns bool, not int) we must preserve unchanged.
+ *
+ * Defined in this file (not its own) so it is always available wherever the core
+ * autoloader has loaded sfDatabaseSessionStorage; it is only referenced from above.
+ */
+class sfDatabaseSessionHandler implements SessionHandlerInterface
+{
+    /** @var sfDatabaseSessionStorage */
+    private $storage;
```
thePanz: we can typehint this property

lib/storage/sfDatabaseSessionStorage.class.php
```diff
@@ -190,3 +185,67 @@ public function regenerate($destroy = false)
         return $this->sessionWrite($newId, $this->sessionRead($currentId));
     }
 }
+
+/**
+ * Adapts an sfDatabaseSessionStorage to the SessionHandlerInterface required by
+ * PHP 8.4's session_set_save_handler().
```
thePanz, line 191:
```suggestion
 * PHP 8.4's session_set_save_handler().
 * See https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#session_set_save_handler
```

lib/storage/sfDatabaseSessionStorage.class.php
JohannesTyra: Done in 0b6cd8c — switched to `private sfDatabaseSessionStorage $storage;` (dropped the now-redundant `@var`). Min PHP is `^7.4` so the typed property is safe.

lib/storage/sfDatabaseSessionStorage.class.php
JohannesTyra: Added the RFC link to the docblock in 0b6cd8c. Thanks!

lib/storage/sfDatabaseSessionStorage.class.php
```diff
@@ -190,3 +185,67 @@ public function regenerate($destroy = false)
...
+ * See https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#session_set_save_handler
+ *
+ * This must be a separate object rather than having sfDatabaseSessionStorage
+ * implement the interface itself: sfSessionStorage already defines read()/write()
+ * for symfony's attribute storage ($_SESSION[$key]), which collide with the
+ * interface's read()/write() save-handler methods. The adapter delegates to the
+ * storage's sessionOpen/sessionClose/sessionRead/sessionWrite/sessionDestroy/sessionGC.
+ *
+ * #[\ReturnTypeWillChange] is required: SessionHandlerInterface declares tentative
```
thePanz: According to https://www.php.net/manual/en/class.sessionhandlerinterface.php the interface has types to the method's parameters.

We should keep them, and make sure that the values we return are aligned to the interface.
This would make the implementation more robust. WDYT?

JohannesTyra: Good call — done in 634a3ce. Added the interface's scalar parameter types and covariant scalar return types (`bool`/`string`/`int`), which lets me drop all the `#[\ReturnTypeWillChange]` attributes.

One constraint: the lib still targets PHP 7.4 (`^7.4 || ^8.1`), and the interface's `read(): string|false` / `gc(): int|false` are union return types (8.0+), so they can't be declared here. I used the covariant `string` / `int` instead — both are subtypes of the tentative types, so no deprecation is emitted on 8.x. `sessionGC()` returns `bool` in all three storage subclasses (PDO/MySQL/PostgreSQL), so `gc()` casts to `int` to honour the interface. Verified on 8.4.21 with `E_ALL`: no deprecation, no TypeError.

JohannesTyra: Tried this (634a3ce) but had to back the method types out in a5ab2a0: the lib still supports PHP 7.4, and on 7.4 the internal `SessionHandlerInterface` has **untyped** parameters. Declaring the interface's parameter/return types on the adapter narrows that contract, which raises an "incompatible declaration" warning and fails `sfPDOSessionStorageTest` on 7.4. CI confirmed it — 8.1–8.5 green, 7.4 red: https://github.com/FriendsOfSymfony1/symfony1/actions/runs/26810096735

The union return types (`read(): string|false`, `gc(): int|false`) can't be declared on 7.4 either (union types are 8.0+). So I've kept the save-handler methods on `#[\ReturnTypeWillChange]` — the only portable option while 7.4 is supported — and left the typed `$storage` property in place. If/when 7.4 support is dropped, the full typing becomes possible and I'm happy to revisit then.

PR #398, opened 2026-07-14
Allow overriding sfYaml class

thePanz:
This will:
1. quote current configuration strings containing `@` symbol, as required by Yaml v1.2
2. split the sfYaml::load() into parse() and parseFile() (aligns with current Symfony Yaml component)
3. allow to override which sfYaml class is being used

This allows to migrate to Symfony Yaml component:
```php
// file: apps/frontend/config/frontendConfiguration.class.php

class frontendConfiguration extends sfApplicationConfiguration
{
    public function configure(): void
    {
        sfConfig::set('sf_yaml_class', MyNewYamlParser::class);
    }
```


lib/addon/sfData.class.php
```diff
@@ -93,7 +93,8 @@ public function getFiles($element = null)
     protected function doLoadDataFromFile($file)
     {
         // import new datas
-        $data = sfYaml::load($file, sfConfig::get('sf_charset', 'UTF-8'));
+
+        $data = sfConfig::getYamlClass()::parseFile($file, sfConfig::get('sf_charset', 'UTF-8'));
```
alquerci, line 97: **suggestion (non-blocking):** `sfYaml` can be a facade that use `sfConfig::getYamlClass()` to select the configured implementation.

That can avoid this tricky expression.

thePanz: Thank you for your comment @alquerci , the proposed solution would alter *all* usages of sfYaml, including usage from an application's code-base, not just from the Symfony1 internals.

While it would make things more compact, and requires less core-changes, the current solution would reading the code explicitly state if we directly use the sfYaml class, or the one defined by the sfConfig.
That would allow a more precise migration of projects willing to ditch the sfYaml usage.

alquerci: **suggestion (non-blocking):** So to keep choice, needs to add a side facade like `sfYamlFacade`.

My concern is to avoid this tricky expression `sfConfig::getYamlClass()::methodName(`.

thirsch: Replacing `X::Y` with a facade would not only eliminate this chained notation but also clearly defining the interface to the Yaml class within our codebase.

But I tend to take a pragmatic approach and think it’s fine. Looks like the interface haven't changed a lot over the years.

thirsch (approved):

PR #399, opened 2026-07-14
Remove executable flag from autoload.php and composer.json files

thePanz:


mentalstring:
There's a few other files with executable-bit, including [images](https://github.com/FriendsOfSymfony1/symfony1/blob/master/data/web/sf/sf_admin/images/cancel.png), and some random .php files (some from sfDoctrinePlugin) that don't have a shebang to run as is.

Should we fix that too, or do you think it can break some setups, @thePanz?

thePanz:
@mentalstring yeah, that's a good idea: we should remove the executable flag from those :+1:

